PR #3, opened 2016-05-13
allow a single node to pass intrapod health checks

sofuture:
A single node cluster should be valid, so as long as this check is hardcoded, it should pass with a single node.


a-palchikov:
@sofuture: The point of this test is checking connectivity between nodes - it does not make sense to have it running with just one. I'm not sure what your use-case is, but maybe disabling this check in case of a single node cluster is a way to go.

sofuture:
"Intra" naming is misleading, I will fix and tackle in a more sensible way.

a-palchikov:
@sofuture thanks!

PR #4, opened 2016-05-16
Rename interpod check, allow it to be disabled

sofuture:
Rename 'intra'->'inter' in name of cross pod communication health checks. Also allow it to be disabled on the agent command line.


a-palchikov:
lgtm

PR #6, opened 2017-02-07
Dmitri/1671 status update

a-palchikov:
* Moved backend activity to the agent's status loop - this is the only place where backends are updated. The updates are serialized to avoid racing
 * Downgraded informational logging from `Info` to `Debug` level
 * Parallelized checker tests
 * Replaced k8s dependencies with the light-weight client package


agent/agent.go
```diff
@@ -192,24 +195,49 @@ func (r *agent) LocalStatus() *pb.NodeStatus {
...
-		c.Check(&reporter)
+		select {
+		case semaphoreCh <- struct{}{}:
+			go func(c health.Checker) {
+				var probes health.Probes
+				log.Debugf("running checker %v", c.Name())
+				c.Check(&probes)
+				probeCh <- probes
+				// release checker slot
+				<-semaphoreCh
```
klizhentas: could crash or hang during check and consume semaphore slot forever. I would  factor out this goroutine, pass cancel context with timeout and add defer release

agent/agent.go
```diff
@@ -192,24 +195,49 @@ func (r *agent) LocalStatus() *pb.NodeStatus {
...
+			}(c)
+		case <-ctx.Done():
+			return nil, trace.Errorf("timed out")
```
klizhentas: `trace.ConnectionProblem` is more appropriate

agent/agent.go
```diff
@@ -256,27 +289,23 @@ func (r *agent) collectStatus(ctx context.Context) (systemStatus *pb.SystemStatu
...
+			go r.getStatusFrom(ctx, member, statusCh)
 		}
 	}
-	wg.Wait()
-	close(statuses)
 
-	for status := range statuses {
-		log.Infof("retrieved status from %v: %v", status.member, status.NodeStatus)
+	for i := 0; i < len(members); i++ {
+		status := <-statusCh
```
klizhentas: should handle case with timeouts as well, right now you will hang forever.

PR #11, opened 2017-03-24
Dmitri/cache

a-palchikov:
Adds in-memory cache that persists the last cluster status.
Also drops sqlite support.


agent/backend/inmemory/inmemory.go
```diff
@@ -0,0 +1,34 @@
+package inmemory
+
+import pb "github.com/gravitational/satellite/agent/proto/agentpb"
+
+// New creates a new instance of cache
+func New() *cache {
```
klizhentas: is it multi-threaded?

klizhentas: could be called by multiple goroutines?

r0mant:
What about influxdb backend? Do we want to keep it?

klizhentas:
cache does not seem to be protected for concurrent access. can you access it via multiple goroutines @a-palchikov

a-palchikov:
I think we should influxdb - even if it's not used currently, it serves as an example of implementing a backend for satellite.

PR #18, opened 2017-07-30
Sasha/ip

klizhentas:
* Add IP v4 check
* Switch to dep for dependency management
* Migrate to kubernetes 2.0.0


build.assets/grpc/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM golang:1.7-wheezy
```
r0mant: Since we're migrating to 1.8.3, use it here too?

monitoring/ip.go
```diff
@@ -0,0 +1,58 @@
+package monitoring
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gravitational/satellite/agent/health"
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+	"github.com/gravitational/trace"
```
r0mant: Add empty line to separate import grops.

monitoring/ip.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Check runs a health check and records any errors into the specified reporter.
+func (s *IPForwardChecker) Check(ctx context.Context, reporter health.Reporter) {
+	value, err := Sysctl(IPForwardParam)
+	if err != nil {
+		reporter.Add(NewProbeFromErr(
+			IPForwardCheck,
+			trace.Wrap(
+				trace.ConvertSystemError(err),
```
a-palchikov: Looks like `ConvertSystemError` should accept a format and arguments just like the other trace error construction APIs:
```go
trace.ConvertSystemError(err, "failed to execute sysctl %v", IPForwardParam)
```

monitoring/sysctl.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// Sysctl returns kernel parameter by reading proc/sys
+func Sysctl(name string) (string, error) {
+	path := filepath.Clean(filepath.Join("proc", "sys", strings.Replace(name, ".", "/", -1)))
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "", trace.ConvertSystemError(err)
+	}
+	return string(data[:len(data)-1]), nil
```
a-palchikov: You should be careful with empty `data` here

monitoring/ip.go
klizhentas: out of scope of this PR

PR #22, opened 2017-10-09
[WIP] New metrics

UnderGreen:


monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+			prometheus.BuildFQName(namespace, "etcd", "leader"),
+			"Endpoint is leader of ETCD cluster.",
+			nil, nil,
+		), prometheus.GaugeValue},
+	}, nil
+}
+
+// Collect implements prometheus.Collector.
+func (e *ETCDCollector) Collect(ch chan<- prometheus.Metric) error {
+	e.mutex.Lock() // To protect metrics from concurrent collects.
```
klizhentas: why protect it? what would happen if used concurrently?

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	if bytes.Contains(resp.Bytes(), []byte("not current leader")) {
+		m, err = e.isLeader.newConstMetric(0.0)
+		if err != nil {
+			success = false
+		}
+		ch <- m
+	} else {
+		health, err := e.healthStatus()
```
klizhentas: health -> healthy

Gopkg.toml
```diff
@@ -83,7 +83,7 @@
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "4.0.0"
+  branch = "release-4.0"
```
a-palchikov: Do you know what the difference is between `4.0.0` and `release-4.0`?

a-palchikov:
I think you should add some details about what changes PR introduces - does not have to be much - but the most important high-level changes

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+
+// PlanetCollector implements the prometheus.Collector interface.
+type PlanetCollector struct {
+	configETCD monitoring.ETCDConfig
+	Collectors map[string]Collector
+}
+
+// NewPlanetCollector creates a new PlanetCollector
+func NewPlanetCollector(configETCD *monitoring.ETCDConfig) (*PlanetCollector, error) {
+	collectors := make(map[string]Collector)
```
a-palchikov: move this after the initialization as it will go unused if `NewETCDCollector` fails

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+	var m prometheus.Metric
+	var err error
+	success := true
+
+	resp, err := e.client.Get(e.client.Endpoint("v2", "stats", "leader"), url.Values{})
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	if bytes.Contains(resp.Bytes(), []byte("not current leader")) {
```
klizhentas: this whole logic block should be refactored to:

```
// this is your default value, returned by default if all other nested blocks failed
metric, err = e.health.newConstMetric(0.0)
if err != nil {
  // nothing to do, return
  return
}

// then refactor logic to emit 1 if etcd is healthy, otherwise return the default metric
healthy, _ := e.healthStatus()
if !bytes.Contains(resp.Bytes(), []byte("not current leader")) && healthy {
   metric, err = e.health.newConstMetric(1.0)
}

ch <- metric

```

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return false, trace.Wrap(err)
+	}
+
+	err = json.Unmarshal(resp.Bytes(), &result)
+	if err != nil {
+		err = json.Unmarshal(resp.Bytes(), &nresult)
+	}
+	if err != nil {
+		return false, trace.Wrap(err, "unable to parse JSON output: %s", resp.Bytes())
```
klizhentas: it will be  a byte stream here, cast to string

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,86 @@
...
+			prometheus.BuildFQName(namespace, "sysctl", "br_netfilter"),
+			"Status of bridge netfilter kernel param",
+			nil, nil,
+		), prometheus.GaugeValue},
+	}
+}
+
+// Collect implements prometheus.Collector.
+func (s *SysctlCollector) Collect(ch chan<- prometheus.Metric) error {
+	s.mutex.Lock() // To protect metrics from concurrent collects.
```
klizhentas: remove this lock, not sure why do you need this

klizhentas: in general, avoid putting locks in the functions that are doing system calls, this is a no-no as you can get a deadlock with this conditions when network hangs

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,86 @@
...
+	defer s.mutex.Unlock()
+
+	var (
+		metric float64
+		m      prometheus.Metric
```
klizhentas: m as a name is not descriptive, call it `metric`

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,86 @@
...
+	)
+
+	param, err := monitoring.Sysctl("net.ipv4.ip_forward")
```
klizhentas: create constants in monitoring package instead of using strings directly

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+	collectorETCD, err := NewETCDCollector(configETCD)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	collectorSysctl := NewSysctlCollector()
```
a-palchikov: why not simply:
```go
collectors["sysctl"] = NewSysctlCollector()
```

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func execute(name string, c Collector, ch chan<- prometheus.Metric) {
+	begin := time.Now()
+	err := c.Collect(ch)
+	duration := time.Since(begin)
+	var success float64
+
+	if err != nil {
+		log.Errorf("%s collector failed after %fs: %s", name, duration.Seconds(), err)
```
a-palchikov: Since the collector runtime is unrestricted, why not use default duration format:
```go
log.Errorf("%s collector failed after %v: %s", name, duration, err)
```

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+		success = 0
+	} else {
+		log.Debugf("%s collector succeeded after %fs.", name, duration.Seconds())
+		success = 1
+	}
+	m, err := prometheus.NewConstMetric(scrapeDurationDesc, prometheus.GaugeValue, duration.Seconds(), name)
+	if err != nil {
+		log.Errorf("failed to collect of collector_duration_seconds metric: %s", err)
```
a-palchikov: `failed to create metric for duration of scrape`
Besides `Errorf` is for hard errors - use `Warnf` if these are informational and non-terminal.

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	ch <- m
```
a-palchikov: Is there a reason to send `m` if `err != nil`?

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+
+	m, err = prometheus.NewConstMetric(scrapeSuccessDesc, prometheus.GaugeValue, success, name)
+	if err != nil {
+		log.Errorf("failed to collect of collector_success metric: %s", err)
+	}
+	ch <- m
```
a-palchikov: Same here

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Collector is the interface a collector has to implement.
```
a-palchikov: This description sounds like a recursion. Can you describe what a collector is?

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+type Collector interface {
+	// Get new metrics and expose them via prometheus registry.
```
a-palchikov: `Collect collects metrics and exposes them to the prometheus registry on the specified channel.
Returns an error if collection fails`

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+	config monitoring.ETCDConfig
+	mutex  sync.RWMutex
+
+	isRunning typedDesc
+	health    typedDesc
+	isLeader  typedDesc
+}
+
+// NewETCDCollector returns an initialized ETCDCollector.
+func NewETCDCollector(config *monitoring.ETCDConfig) (Collector, error) {
```
a-palchikov: You should return the concrete type (i.e. `*ETCDCollector` in this case) - this will to call additional methods on the returned object not exposed by the interface if necessary.

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+	transport, err := config.NewHTTPTransport()
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	if len(config.Endpoints) == 0 {
```
a-palchikov: Validate input before doing anything else - i.e. move this check to the top

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+			nil, nil,
+		), prometheus.GaugeValue},
+		health: typedDesc{prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "etcd", "health"),
+			"Health status of ETCD.",
+			nil, nil,
+		), prometheus.GaugeValue},
+		isLeader: typedDesc{prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "etcd", "leader"),
+			"Endpoint is leader of ETCD cluster.",
```
a-palchikov: `... is the leader ...`

monitoring/collector/etcd.go
a-palchikov: Embed the mutex:
```go
type ETCDCollector struct {
	sync.RWMutex
}
// use a short-hand access
var collector *ETCDCollector
collector.Lock()
```
as this apparently affects all attributes

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+	}, nil
+}
+
+// Collect implements prometheus.Collector.
+func (e *ETCDCollector) Collect(ch chan<- prometheus.Metric) error {
+	e.mutex.Lock() // To protect metrics from concurrent collects.
+	defer e.mutex.Unlock()
+
+	var m prometheus.Metric
+	var err error
```
a-palchikov: this declaration is redundant

monitoring/collector/etcd.go
a-palchikov: Can you factor out this logic?
```go
if !isLeaderResponse(resp) {
```

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+
+	resp, err := e.client.Get(e.client.Endpoint("v2", "stats", "leader"), url.Values{})
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	if bytes.Contains(resp.Bytes(), []byte("not current leader")) {
+		m, err = e.isLeader.newConstMetric(0.0)
+		if err != nil {
+			success = false
```
a-palchikov: Handling errors like this makes me wonder if `m` is useful when `err != nil`? Can you elaborate?

monitoring/collector/etcd.go
a-palchikov: `err` is not handled - you should probably hint at what it was in the log

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		ch <- m
+	} else {
+		health, err := e.healthStatus()
+		if err != nil {
+			success = false
+		}
+		if health {
+			m, err = e.health.newConstMetric(1.0)
```
a-palchikov: these blocks are identical bar the float constant - is there a way to avoid duplication?

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+		if err != nil {
+			success = false
+		}
+		ch <- m
+	}
+
+	if !success {
+		m, err = e.isRunning.newConstMetric(0.0)
+		if err != nil {
+			success = false
```
a-palchikov: redundant - `success` is already `false`

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"strconv"
+	"sync"
+
+	"github.com/gravitational/satellite/monitoring"
+	"github.com/gravitational/trace"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// SysctlCollector collects metrics from kernel system params
```
a-palchikov: ` ... converts kernel system parameters to prometheus metrics`

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,86 @@
...
+type SysctlCollector struct {
+	ipv4Forwarding typedDesc
+	brNetfilter    typedDesc
+	mutex          sync.RWMutex
+}
+
+// NewSysctlCollector returns an initialized SysctlCollector.
+func NewSysctlCollector() Collector {
```
a-palchikov: same here - strive to return a concrete type - this has no effect on external callers

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,86 @@
...
+	return &SysctlCollector{
+		ipv4Forwarding: typedDesc{prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "sysctl", "ipv4_forwarding"),
+			"Status of IPv4 forwarding kernel param",
```
a-palchikov: `Value of IPv4 forwarding kernel parameter`

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,86 @@
...
+			nil, nil,
+		), prometheus.GaugeValue},
+		brNetfilter: typedDesc{prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "sysctl", "br_netfilter"),
+			"Status of bridge netfilter kernel param",
```
a-palchikov: `Value of bridge netfilter module parameter`

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,86 @@
...
+
+	param, err := monitoring.Sysctl("net.ipv4.ip_forward")
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if metric, err = strconv.ParseFloat(param, 64); err != nil {
+		return trace.Wrap(err)
+	}
+	if m, err = s.ipv4Forwarding.newConstMetric(metric); err != nil {
+		return trace.Wrap(err)
```
a-palchikov: Here, you're handling failure to create a metric as an error (in contrast to all other places)

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	if len(config.Endpoints) == 0 {
+		return nil, trace.BadParameter("no ETCD endpoints configured")
+	}
+
+	client, err := newClient(config.Endpoints[0], roundtrip.HTTPClient(&http.Client{
+		Transport: transport,
+		Timeout:   collectMetricsTimeout,
+	}))
```
r0mant: err is ignored

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+		"planet_exporter: Whether a collector succeeded.",
+		[]string{"collector"},
+		nil,
+	)
+)
+
+// PlanetCollector implements the prometheus.Collector interface.
+type PlanetCollector struct {
+	configETCD monitoring.ETCDConfig
+	Collectors map[string]Collector
```
r0mant: Godoc on exported field please.

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Collect implements the prometheus.Collector interface.
+func (pc *PlanetCollector) Collect(ch chan<- prometheus.Metric) {
+	wg := sync.WaitGroup{}
+	wg.Add(len(pc.Collectors))
+	for name, c := range pc.Collectors {
+		go func(name string, c Collector) {
+			execute(name, c, ch)
+			wg.Done()
```
r0mant: I would do `defer wg.Done()` at the beginning of the function instead. This way it'll be called even if "execute" panics and generally is a good practice.

monitoring/collector/collector.go
r0mant: trace.DebugReport(err)

monitoring/collector/collector.go
r0mant: trace.DebugReport(err)

monitoring/collector/collector.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	m, err := prometheus.NewConstMetric(scrapeDurationDesc, prometheus.GaugeValue, duration.Seconds(), name)
+	if err != nil {
+		log.Errorf("failed to collect of collector_duration_seconds metric: %s", err)
+	}
+	ch <- m
+
+	m, err = prometheus.NewConstMetric(scrapeSuccessDesc, prometheus.GaugeValue, success, name)
+	if err != nil {
+		log.Errorf("failed to collect of collector_success metric: %s", err)
```
r0mant: trace.DebugReport(err)

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return &ETCDCollector{
+		client: client,
+		config: *config,
+		isRunning: typedDesc{prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "etcd", "up"),
+			"Whether scraping ETCD metrics was successful.",
```
r0mant: ETCD reads weird, can you replace with "etcd" everywhere? It's not an abbreviation.

monitoring/collector/etcd.go
```diff
@@ -0,0 +1,169 @@
...
+		if health {
+			m, err = e.health.newConstMetric(1.0)
+			if err != nil {
+				success = false
+			}
+			ch <- m
+		} else {
+			m, err = e.health.newConstMetric(0.0)
+			if err != nil {
+				success = false
```
r0mant: err is not logged anywhere in these blocks. I think we should log it.

Gopkg.toml
UnderGreen: I changed it when tried to use `dep ensure` in planet repo. But I still cannot fix dependency hell, caused by renaming Sirupsen -> sirupsen. After all fixes build of planet crashes on runc library.

monitoring/collector/collector.go
UnderGreen: Good point.

monitoring/collector/etcd.go
UnderGreen: I can't use default metric here, because it is a constant metric. Prometheus documentation says about using constant metrics for situation when you should scrape them from outside. I can switch to non-constant metric and then code with defaults will work(what I did before, but then switched to constant metrics).

monitoring/collector/collector.go
UnderGreen: Fixed

monitoring/collector/collector.go
UnderGreen: Fixed too

monitoring/collector/sysctl.go
UnderGreen: Refactored etcd collect method, for now there is normal handling of errors

monitoring/collector/etcd.go
a-palchikov: I feel there's too much noise around using constant metric values - including the metrics constructor giving an error back for a precondition test which needs to be checked on every metric generation.
From what I see, there're just two types of constant metrics used: `off` and `on` (subject to interpretation).
I'd rather have them declared as package locals:

[collector/etcd.go]
```go
// ETCDCollector collects ETCD stats from the given server and exports them using
// the prometheus metrics package.
type ETCDCollector struct {
	client *roundtrip.Client
	config monitoring.ETCDConfig
}

// NewETCDCollector returns an initialized ETCDCollector.
func NewETCDCollector(config *monitoring.ETCDConfig) (Collector, error) {
	transport, err := config.NewHTTPTransport()
	if err != nil {
		return nil, trace.Wrap(err)
	}

	if len(config.Endpoints) == 0 {
		return nil, trace.BadParameter("no ETCD endpoints configured")
	}

	client, err := newClient(config.Endpoints[0], roundtrip.HTTPClient(&http.Client{
		Transport: transport,
		Timeout:   collectMetricsTimeout,
	}))

	return &ETCDCollector{
		client: client,
		config: *config,
	}, nil
}

// Collect implements prometheus.Collector.
func (e *ETCDCollector) Collect(ch chan<- prometheus.Metric) error {
	resp, err := e.client.Get(e.client.Endpoint("v2", "stats", "leader"), url.Values{})
	if err != nil {
		return trace.Wrap(err)
	}

	if !isLeaderResp(resp.Bytes()) {
		ch <- leaderOff
		ch <- runningOn
		return nil
	}

	healthy, err := e.healthStatus()
	if err != nil {
		return trace.Wrap(err)
	}

	ch <- leaderOn
	if healthy {
		ch <- healthyOn
	} else {
		ch <- healthyOff
	}

	ch <- runningOn

	return nil
}

func isLeaderResp(resp []byte) bool {
	return bytes.Contains(resp, []byte("not current leader"))
}

var isRunning = typedDesc{prometheus.NewDesc(
	prometheus.BuildFQName(namespace, "etcd", "up"),
	"Whether scraping ETCD metrics was successful.",
	nil, nil,
), prometheus.GaugeValue}

var runningOn = isRunning.mustNewConstMetric(1.0)

var healthy = typedDesc{prometheus.NewDesc(
	prometheus.BuildFQName(namespace, "etcd", "health"),
	"Health status of ETCD.",
	nil, nil,
), prometheus.GaugeValue}

var healthyOn = healthy.mustNewConstMetric(1.0)
var healthyOff = healthy.mustNewConstMetric(0.0)

var isLeader = typedDesc{prometheus.NewDesc(
	prometheus.BuildFQName(namespace, "etcd", "leader"),
	"Endpoint is leader of ETCD cluster.",
	nil, nil,
), prometheus.GaugeValue}

var leaderOn = isLeader.mustNewConstMetric(1.0)
var leaderOff = isLeader.mustNewConstMetric(0.0)
```

monitoring/collector/sysctl.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Collect implements prometheus.Collector.
+func (s *SysctlCollector) Collect(ch chan<- prometheus.Metric) error {
+	var (
+		parsedMetric float64
+		metric       prometheus.Metric
+	)
+
+	param, err := monitoring.Sysctl(IPv4Forwarding)
```
a-palchikov: It would get repetitive real quick when we start adding sysctl parameters to collect:
```go
func (s *SysctlCollector) Collect(ch chan<- prometheus.Metric) error {
	...
	metric, err := sysctlMetric(IPv4Forwarding, s.ipv4Forwarding)
	if err != nil {
		return trace.Wrap(err)
	}
	ch <- metric
	...
}

func sysctlMetric(paramName string, typedDesc desc) (prometheus.Metric, error) {
	param, err = monitoring.Sysctl(paramName)
	if err != nil {
		return nil, trace.Wrap(err)
	}
	if parsedMetric, err = strconv.ParseFloat(param, 64); err != nil {
		return nil, trace.Wrap(err)
	}
	if metric, err = desc.newConstMetric(parsedMetric); err != nil {
		return nil, trace.Wrap(err)
	}
	return metric, nil
}
```

klizhentas:
@UnderGreen please split your work in logical items, e.g. Kuberntes collector should be a separate PR and docker collector should be a separate PR, and this one should be merged.

UnderGreen:
Then I close this PR and going to create new PR's.

PR #27, opened 2017-10-12
Kubernetes metrics collector

UnderGreen:


monitoring/collector/kubernetes.go
```diff
@@ -0,0 +1,88 @@
...
+		FieldSelector: fields.Everything().String(),
+	}
+	nodes, err := k.client.Nodes().List(listOptions)
+	if err != nil {
+		return trace.Wrap(err, "failed to query nodes: %v", err)
+	}
+
+	for _, item := range nodes.Items {
+		for _, condition := range item.Status.Conditions {
+			log.Debugf("node item: %v", item)
```
a-palchikov: this will contain a lot of almost unreadable data - do you want to dump only relevant attributes?

monitoring/collector/kubernetes.go
```diff
@@ -0,0 +1,88 @@
...
+			if condition.Type != v1.NodeReady {
+				continue
```
a-palchikov: What about other states - `outofdisk`, `memorypressure`, `diskpressure` and `inodepressure`?

PR #28, opened 2017-10-12
Docker metrics collector

UnderGreen:


monitoring/collector/docker.go
```diff
@@ -0,0 +1,93 @@
...
+		ch <- d.dockerUp.mustNewConstMetric(1.0)
+		return nil
+	}
+
+	ch <- d.dockerUp.mustNewConstMetric(0.0)
+	return nil
+}
+
+// healthStatus determines status of docker service
+// by checking url http://docker/version from daemon socket
```
a-palchikov: `by fetching the docker's version HTTP endpoint`

monitoring/collector/docker.go
```diff
@@ -0,0 +1,93 @@
...
+func (d *DockerCollector) healtStatus() (bool, error) {
```
a-palchikov: healt**h**Status

monitoring/collector/docker.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Collect implements the prometheus.Collector interface
+func (d *DockerCollector) Collect(ch chan<- prometheus.Metric) error {
+	healthy, err := d.healtStatus()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if healthy {
+		ch <- d.dockerUp.mustNewConstMetric(1.0)
```
a-palchikov: I don't think it's a good idea to invoke potentially panicking methods in a normal flow unless there's no altrernative - do this as soon as possible (preferably on startup), so factor `dockerUp.mustNewConstMetric` to `dockerOn` and `dockerOff` package locals or something similar.

monitoring/collector/docker.go
```diff
@@ -0,0 +1,93 @@
...
+		client: client,
+		dockerUp: typedDesc{prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "docker", "health"),
+			"Status of Docker daemon",
+			nil, nil,
+		), prometheus.GaugeValue},
+	}, nil
+}
+
+// Collect implements the prometheus.Collector interface
```
a-palchikov: I've been looking at prometheus docs and the Collector interface does not have this method (it has one w/o the error return) - where is this used or translated to prometheus Collector interface?

UnderGreen: https://godoc.org/github.com/prometheus/client_golang/prometheus#Collector

UnderGreen: Hmm, really without error return.

monitoring/collector/collector.go
```diff
@@ -50,14 +50,14 @@ var (
 	)
 )
 
-// PlanetCollector implements the prometheus.Collector interface.
-type PlanetCollector struct {
+// MetricsCollector implements the prometheus.Collector interface.
```
a-palchikov: Can you add that the metrics collector is a collection of other collector implementations

monitoring/collector/collector.go
```diff
@@ -128,6 +128,8 @@ func execute(name string, c Collector, ch chan<- prometheus.Metric) {
 	}
 }
 
+// Collector is the interface implemented by anything that can be
+// used by Prometheus to collect metrics.
```
a-palchikov:
```
Collector defines an interface for collecting specific facts about a system and expose them to
prometheus on the provided channel as metrics.
```

monitoring/collector/docker.go
```diff
@@ -0,0 +1,100 @@
...
+func (d *DockerCollector) Collect(ch chan<- prometheus.Metric) error {
+	var metric prometheus.Metric
+
+	healthy, err := d.healthStatus()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if healthy {
+		if metric, err = d.dockerUp.newConstMetric(1.0); err != nil {
+			return trace.Wrap(err, "failed to create prometheus metric: %v", err)
```
a-palchikov: i'd still get rid of this error handling if I could help it. It will almost never trigger (as long as it validates the cardinality of label sets) and moving this to init phase will make this code much cleaner

PR #31, opened 2017-10-12
boot config params checks

DenisMishin:
partially implements https://github.com/gravitational/gravity/issues/2843 and will be used to check storage / network driver manifest requirements

### DIFF
--- a/cmd/agent/local.go
+++ b/cmd/agent/local.go
@@ -8,11 +8,12 @@ import (
 	"github.com/gravitational/satellite/monitoring"
 )
 
-func localChecks() {
+func localChecks() error {
 	ch := monitoring.BasicCheckers()
 	var r health.Probes
 
 	ch.Check(context.TODO(), &r)
 
 	fmt.Printf("%+v\n", r)
+	return nil
 }
--- a/monitoring/boot_config.go
+++ b/monitoring/boot_config.go
@@ -1,0 +1,186 @@
+/*
+Copyright 2016 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package monitoring
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"regexp"
+	"syscall"
+
+	"github.com/gravitational/satellite/agent/health"
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+	"github.com/gravitational/trace"
+)
+
+func int8string(bs []int8) string {
+	b := make([]byte, 0, len(bs))
+	for _, v := range bs {
+		if v == 0 {
+			break
+		}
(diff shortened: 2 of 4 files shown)


cmd/agent/local.go
```diff
@@ -8,11 +8,12 @@ import (
 	"github.com/gravitational/satellite/monitoring"
 )
 
-func localChecks() {
+func localChecks() error {
```
a-palchikov: why this since it never returns an error - in preparation of future changes or to make it look symmetrical to `status`?

monitoring/boot_config.go
```diff
@@ -0,0 +1,186 @@
...
+	file, err := os.Open(fmt.Sprintf("/boot/config-%s", int8string(uname.Release[:])))
+	if err != nil {
+		return nil, trace.ConvertSystemError(err)
+	}
+	defer file.Close()
+
+	params := map[string]string{}
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if reBootConfigComment.Match(scanner.Bytes()) || scanner.Text() == "" {
```
a-palchikov: does this have to be a regexp for comments - you could simply check for `#` in front by skipping whitespace and the condition is better reversed: if-empty then match

monitoring/boot_config.go
```diff
@@ -0,0 +1,186 @@
...
+// Name returns name of the checker
+func (c *BootConfigParamChecker) Name() string {
+	return bootConfigParamID
+}
+
+// Check parses boot config files and validates whether parameters provided are set
+func (c *BootConfigParamChecker) Check(ctx context.Context, reporter health.Reporter) {
+	cfg, err := parseBootConfig()
+	if trace.IsNotFound(err) {
+		reporter.Add(NewProbeFromErr(bootConfigParamID,
```
a-palchikov: This is not an error, right?

monitoring/boot_config.go
```diff
@@ -0,0 +1,186 @@
...
+			BootConfigParam{"CONFIG_PID_NS"},
+			BootConfigParam{"CONFIG_IPC_NS"},
+			BootConfigParam{"CONFIG_UTS_NS"},
+			BootConfigParam{"CONFIG_CGROUPS"},
+			BootConfigParam{"CONFIG_CGROUP_CPUACCT"},
+			BootConfigParam{"CONFIG_CGROUP_DEVICE"},
+			BootConfigParam{"CONFIG_CGROUP_FREEZER"},
+			BootConfigParam{"CONFIG_CGROUP_SCHED"},
+			BootConfigParam{"CONFIG_CPUSETS"},
+			BootConfigParam{"CONFIG_MEMCG"},
```
a-palchikov: I think this is not enough - although the kernel is configured with memory cgroup support, it can still be turned off in the boot config in `/boot/grub/grub.cfg` (see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/resource_management_guide/sec-memory).
I think the only confident way is checking cgroups is `/proc/cgroups` or cgroup mounts in `/proc/mounts`

cmd/agent/local.go
DenisMishin: yes, consistency

monitoring/boot_config.go
DenisMishin: that's for a side effect to log the underlying error where you could see filename etc when invoked as part of gravity checks

monitoring/boot_config.go
DenisMishin: yes, that's separate checker though

cmd/agent/local.go
a-palchikov: i mean, why invent the error if one is not there?..

PR #32, opened 2017-10-13
Use apiserver service

UnderGreen:


a-palchikov:
Does not look to be necessary, so closing.

PR #35, opened 2017-10-27
Configure TLS for HTTPS/gRPC with CA certificate.

a-palchikov:
Had to make some changes to build proper TLS config with CA certificate since otherwise multi-node agent do not connect to each other.


a-palchikov:
will re-test under new conditions

PR #39, opened 2017-11-23
Add a checker for loaded kernel modules.

a-palchikov:
The checker is required, for instance, to ensure that `ebtables` module (or any of it's dependencies) are loaded.


monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+func NewModuleChecker(names ...string) ModuleChecker {
+	return ModuleChecker{
+		Modules:      names,
+		moduleGetter: moduleGetterFunc(LoadModules),
+	}
+}
+
+// Name returns name of the checker
+func (r ModuleChecker) Name() string {
+	return "module-check"
```
klizhentas: better name would be `kernel-module`

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+
+	"github.com/gravitational/trace"
+	log "github.com/sirupsen/logrus"
+)
+
+// NewModuleChecker creates a new kernel module checker
+func NewModuleChecker(names ...string) ModuleChecker {
+	return ModuleChecker{
+		Modules:      names,
+		moduleGetter: moduleGetterFunc(LoadModules),
```
klizhentas: I would rename to `getModule`

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+	"strings"
+
+	"github.com/gravitational/satellite/agent/health"
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+
+	"github.com/gravitational/trace"
+	log "github.com/sirupsen/logrus"
+)
+
+// NewModuleChecker creates a new kernel module checker
```
klizhentas: NewKernelModuleChecker because module is ambiguous name

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+func (r ModuleChecker) check(ctx context.Context, reporter health.Reporter) error {
+	modules, err := r.moduleGetter.Modules()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	var errors []error
+	for _, module := range r.Modules {
+		if !modules.HasLoaded(module) {
+			errors = append(errors, trace.NotFound("module %q not loaded", module))
```
klizhentas: kernel module %q is not loaded

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (r ModuleChecker) check(ctx context.Context, reporter health.Reporter) error {
+	modules, err := r.moduleGetter.Modules()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	var errors []error
+	for _, module := range r.Modules {
+		if !modules.HasLoaded(module) {
```
klizhentas: `WasLoaded` as it does not load itself  but is loaded by kernel

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+			errors = append(errors, trace.NotFound("module %q not loaded", module))
+		}
+	}
+
+	if len(errors) == 1 {
```
klizhentas: this logic is already present in `trace.NewAggregate`

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+		return trace.Wrap(errors[0])
+	}
+	return trace.NewAggregate(errors...)
+}
+
+// ModuleChecker checks if the specified set of kernel modules ara loaded
```
klizhentas: are

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+
+// LoadModules loads list of kernel modules from /proc/modules
+func LoadModules() (Modules, error) {
+	f, err := os.Open("/proc/modules")
+	if err != nil {
+		return nil, trace.ConvertSystemError(err)
+	}
+	defer f.Close()
+
+	modules, err := LoadModulesFrom(f)
```
klizhentas: ReadModulesFrom you are not loading them actually

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return modules, nil
+}
+
+// HasLoaded determines whether module names are loaded.
+// names should not have duplicates
+func (r Modules) HasLoaded(names ...string) bool {
+L:
+	for _, module := range r {
```
klizhentas: if Modules were a map this check will be trivial, also why re-implement loop here? WasLoaded(name) will work as expected and users can loop around when necessary to check the list. This function is doing too much IMHO

monitoring/modules.go
```diff
@@ -0,0 +1,209 @@
...
+			moduleS)
+	}
+
+	instanceS := columns[2]
+	instances, err := strconv.ParseInt(instanceS, 10, 32)
+	if err != nil {
+		return nil, trace.BadParameter("invalid instances field: expected integer, but got %q", instanceS)
+	}
+
+	var dependencies []string
```
klizhentas: you are not really using dependencies anywhere, why implement the code?

monitoring/modules_test.go
```diff
@@ -0,0 +1,105 @@
...
+		checker.Check(context.TODO(), &reporter)
+		c.Assert(reporter, test.DeepCompare, testCase.probes, Commentf(testCase.comment))
+	}
+}
+
+func moduleReader() (Modules, error) {
+	return LoadModulesFrom(bytes.NewReader(modulesData))
+}
+
+var modulesData = []byte(`br_netfilter 22209 0 - Live 0xffffffffc063f000
```
klizhentas: add unhappy case with corrupted data plz

monitoring/modules.go
a-palchikov: Unfortunately not but I can fix it in trace

monitoring/modules.go
a-palchikov: If this is used as a library (and not restricted to use by just the kernel module checker), it may be useful to callers - it is just a speculative implementation.

klizhentas: so the answer is we don't need it, so why keep this code :)

PR #40, opened 2017-11-28
Add check for fs.may_detach_mounts not being set

knisbet:
We encountered an issue on RHEL 7.4 based kernels where pods could get stuck in the terminating state. It appears this is related to the following docker issue: https://github.com/moby/moby/issues/22260

Docker has decided to set the sysctl on startup, but older versions of docker will not do this. I've also added a directive to not report an error if the sysctl option isn't found. Based on the comments in https://github.com/moby/moby/pull/34886/files not all kernels will have this particular sysctl key, so we don't want to report errors if the key isn't on the particular kernel version which won't have this issue.

Cheers,

### DIFF
--- a/monitoring/sysctl_checkers.go
+++ b/monitoring/sysctl_checkers.go
@@ -37,3 +37,18 @@ func NewBridgeNetfilterChecker() *SysctlChecker {
 		OnValueMismatch: "kubernetes requires net.bridge.bridge-nf-call-iptables sysctl set to 1, https://www.gravitational.com/docs/faq/#bridge-driver",
 	}
 }
+
+// NewMayDetachMountsChecker checks if fs.may_detach_mounts is set
+// On RHEL 7.4 based kernels, device removals may fail with "device or resource busy" if fs.may_detach_mounts isn't set
+// Under kubernetes this can cause pods to get stuck in the terminating state
+// Docker issue: https://github.com/moby/moby/issues/22260
+// Docker will be setting the option on startup: https://github.com/moby/moby/pull/34886/files
+func NewMayDetachMountsChecker() *SysctlChecker {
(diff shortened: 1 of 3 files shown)


monitoring/sysctl_checkers.go
```diff
@@ -37,3 +37,18 @@ func NewBridgeNetfilterChecker() *SysctlChecker {
...
+// On RHEL 7.4 based kernels, device removals may fail with "device or resource busy" if fs.may_detach_mounts isn't set
+// Under kubernetes this can cause pods to get stuck in the terminating state
+// Docker issue: https://github.com/moby/moby/issues/22260
+// Docker will be setting the option on startup: https://github.com/moby/moby/pull/34886/files
+func NewMayDetachMountsChecker() *SysctlChecker {
+	return &SysctlChecker{
+		CheckerName:     "may-detach-mounts",
+		Param:           "fs.may_detach_mounts",
+		Expected:        "1",
+		OnValueMismatch: "On RHEL 7.4 or newer based kernels, fs.may_detach_mounts should be set to 1 or pods may get stuck in the Terminating state",
```
klizhentas: I would not mention RHEL specifically as we support many others. The best would be to update the FAQ and add a link here, see the example above with net_bridge

knisbet: I'll add a FAQ link.

However, the reason I mentioned RHEL based kernels, is from the sounds of it may_detach_mounts if an extension only present in RHEL based kernels. 
https://bugzilla.redhat.com/show_bug.cgi?id=1441737

At least I haven't been able to find any reference to non RHEL kernels that have this particular setting.

klizhentas: ah, sounds fair then

PR #44, opened 2018-02-13
Fix boot configuration files to build an empty stub on darwin.

a-palchikov:
Otherwise telekube does not buld on darwin.

### DIFF
--- a/monitoring/boot_config_linux.go
+++ b/monitoring/boot_config_linux.go
@@ -1,0 +1,186 @@
+/*
+Copyright 2016 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


monitoring/boot_config_linux.go
```diff
@@ -0,0 +1,186 @@
...
+		if v == 0 {
+			break
+		}
+		b = append(b, byte(v))
+	}
+	return string(b)
+}
+
+var (
+	reBootConfigParam   = regexp.MustCompile(`(\S+)\=([ym])`)
```
knisbet: I know this is copy/paste code, but is the regex supposed to match 'y' and 'm' or 'y' and 'n' (yes/no)?

a-palchikov: This refers to a y/n/m, where `m` stands for `module`

PR #46, opened 2018-02-20
Update OSChecker to match behavior of telekube

a-palchikov:
Update OSChecker to match behavior of the telekube OS distribution checker:

  - read both release and specific version files (if available)
  - match version on prefix

Fixes https://github.com/gravitational/gravity/issues/3145.


monitoring/os_linux.go
```diff
@@ -0,0 +1,353 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/gravitational/satellite/agent/health"
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+	log "github.com/sirupsen/logrus"
```
UnderGreen: I think this line should be below.

monitoring/os_linux.go
```diff
@@ -0,0 +1,353 @@
...
+
+// versionsMatch tests if test is equivalent to info
+func versionsMatch(test, info OSRelease) bool {
+	if strings.ToLower(test.ID) != strings.ToLower(info.ID) {
+		return false
+	}
+
+	// Versions are matched as prefixes, e.g. if a required version is 7.2 then
+	// it matches 7.2, 7.2.1, 7.2.2, etc.
+	if !strings.HasPrefix(info.VersionID, test.VersionID) {
```
knisbet: I have a tiny concern that this could false positive (7.1 would match 7.10), but I can't think of any mainstream linux distro which would trigger the false positive. Might be worth keeping in mind though.

monitoring/os_linux_test.go
```diff
@@ -0,0 +1,161 @@
...
+	}{
+		{
+			releases:   staticOSReleases("ubuntu", "16.04.3", "centos", "7.2"),
+			getRelease: testGetOSRelease(OSRelease{ID: "CentOS", VersionID: "7.2"}),
+			probes:     health.Probes{&pb.Probe{Checker: osCheckerID, Status: pb.Probe_Running}},
+			comment:    "requirements match",
+		},
+		{
+			releases:   staticOSReleases("ubuntu", "16", "centos", "7.2"),
+			getRelease: testGetOSRelease(OSRelease{ID: "ubuntu", VersionID: "16.04.3"}),
```
knisbet: I'm sure this is rooted somewhere else, but I find this sort of interesting. Centos/Redhat are including sub releases (7.2) but ubuntu is only matching the year of the release (16)? For example, at a certain point in time ubuntu 16.04 would be LTS, but you would also have 16.10 version.

a-palchikov: I think the test intention here is to verify that if the requirement is just the release number (`16` for Ubuntu, regardless of the meaning), then we would match any `16.x` version of Ubuntu.
If you see any concerns regarding this simple heuristic - would be good to hear since I did not invest that much time understanding various versioning schemes.

knisbet: Yea, thinking about it a little bit more, I don't see any really bad side effects by just string matching the prefix. 

I think where I probably got hung up, was it wasn't clear to me OSRelease.VersionID wasn't referring to a specific version, but is more of a string match pattern. So maybe consider renaming OSRelease.VersionID to something like VersionPrefix would make it more clear that this doesn't represent a specific version, but is more of a string match.

a-palchikov: Yes, you're right. I see this coming from the fact that the requirements are represented by the same structure as the release information retrieved from the release/version/lsb_release where the semantics of the attributes was used from.

PR #48, opened 2018-02-23
Unexport checker types that require non-trivial initialization

a-palchikov:
This will avoid mistakes when checkers are initialized directly and panic when used.

### DIFF
--- a/monitoring/boot_config_linux.go
+++ b/monitoring/boot_config_linux.go
@@ -34,16 +34,16 @@ import (
 
 // NewBootConfigParamChecker returns a new checker that verifies
 // kernel configuration options
-func NewBootConfigParamChecker(params ...BootConfigParam) *BootConfigParamChecker {
-	return &BootConfigParamChecker{
+func NewBootConfigParamChecker(params ...BootConfigParam) health.Checker {
+	return &bootConfigParamChecker{
 		Params:              params,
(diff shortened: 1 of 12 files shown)


monitoring/boot_config_linux.go
```diff
@@ -34,16 +34,16 @@ import (
 
 // NewBootConfigParamChecker returns a new checker that verifies
 // kernel configuration options
-func NewBootConfigParamChecker(params ...BootConfigParam) *BootConfigParamChecker {
-	return &BootConfigParamChecker{
+func NewBootConfigParamChecker(params ...BootConfigParam) *bootConfigParamChecker {
```
knisbet: So apparently this pattern is considered a linting error "exported func NewBootConfigParamChecker returns unexported type *monitoring.bootConfigParamChecker, which can be annoying to use" https://github.com/golang/lint/issues/210

I know go has the whole accept interfaces return structs thing, but I wonder if returning an interface would be better since it looks like the checkers are written in a fairly consistent fashion?

a-palchikov: This will fail miserably as I'm actually using the concrete in tests to override the attributes.
I know I can start tucking all that in some sort of configuration but I frankly see no point.
This is also news to me regarding the linting error (and I kind of disagree), but well.

PR #50, opened 2018-02-26
Add empty stubs for darwin support

a-palchikov:


monitoring/os.go
```diff
@@ -0,0 +1,35 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package monitoring
+
+// NewOSChecker returns a new checker to verify OS distribution
+// against the list of supported releases.
+//
+// The checker only supports Linux.
+func NewDTypeChecker(path string) noopChecker {
```
r0mant: Should it be OSChecker or something?

a-palchikov: yes, a big typo.

PR #51, opened 2018-03-07
Update checkers to only report errors on successfully accumulated facts

a-palchikov:
- Export `GetOSRelease` API to be able to use the same OS distribution lookup logic in telekube
 - Revisit checkers to only report errors if information has been actually collected.
In some cases (like missing files, etc.) the test is skipped to avoid dubious false positives.
 - Add more (failing) test scenarios


agent/health/health.go
```diff
@@ -49,20 +49,32 @@ type Reporter interface {
...
 	GetProbes() []*pb.Probe
+	// NumProbes returns the number of probes this reporter contains
+	NumProbes() int
 }
 
 // Probes is a list of probes.
 // It implements the Reporter interface.
 type Probes []*pb.Probe
 
+// Add adds an health probe for a specific node.
```
r0mant: an -> a

monitoring/cgroup.go
```diff
@@ -54,10 +54,21 @@ func (r cgroupChecker) Name() string {
 // Check verifies existence of cgroup mounts given in r.cgroups.
 // Implements health.Checker
 func (r cgroupChecker) Check(ctx context.Context, reporter health.Reporter) {
+	err := r.check(ctx, reporter)
+	if err != nil && !trace.IsNotFound(err) {
+		reporter.Add(NewProbeFromErr(r.Name(), "failed to validate cgroup mounts", err))
+		return
+	}
+	if reporter.NumProbes() != 0 {
```
r0mant: Shouldn't this go before `r.check`, right at the beginning of the function? To quit early and avoid cases like the one we saw yesterday when an empty probe resulted in a failed check. Just to be on a safe side.

monitoring/fs_linux.go
```diff
@@ -48,22 +48,33 @@ func (r dtypeChecker) Name() string {
 
 // Check determines if the filesystem mounted on r supports d_type
 func (r dtypeChecker) Check(ctx context.Context, reporter health.Reporter) {
+	if err := r.check(ctx, reporter); err != nil {
+		reporter.Add(NewProbeFromErr(r.Name(),
+			"failed to determine d_type support in filesystem", err))
+		return
+	}
+	if reporter.NumProbes() != 0 {
```
r0mant: Same as above, I would move this to the top of the function.

r0mant: And in all checkers below too.

monitoring/host.go
```diff
@@ -29,58 +29,93 @@ import (
...
 
-	if runtime.NumCPU() < c.MinCPU {
-		failed = true
+	if c.getCPU() < c.MinCPU {
 		reporter.Add(&pb.Probe{
-			Checker: hostCheckerID,
-			Detail:  fmt.Sprintf("need minimum %d CPU, have %d", c.MinCPU, runtime.NumCPU()),
-			Status:  pb.Probe_Failed,
+			Checker: c.Name(),
+			Detail: fmt.Sprintf("at least %d CPUs required, only %d are available",
```
r0mant: If this is a 1-core machine, the message will be grammatically incorrect, "only 1 are available". I would rephrase to avoid is/are, e.g. "have only X".

PR #54, opened 2018-03-26
Add systemd units status collector

UnderGreen:
Ouch.. the latest version of dep do prune automatically.


monitoring/collector/systemd.go
```diff
@@ -0,0 +1,130 @@
...
+	"docker.service":                  1,
+	"etcd.service":                    1,
+	"kube-apiserver.service":          1,
+	"kube-scheduler.service":          1,
+	"kube-kubelet.service":            1,
+	"kube-proxy.service":              1,
+	"kube-controller-manager.service": 1,
+}
+
+type SystemdCollector struct {
```
a-palchikov: Should document the struct

monitoring/collector/systemd.go
```diff
@@ -0,0 +1,130 @@
...
+	dbusConn         *dbus.Conn
+	systemdState     typedDesc
+	systemdUnitState typedDesc
+}
+
+func NewSystemdCollector() (*SystemdCollector, error) {
```
a-palchikov: Should document the export

monitoring/collector/systemd.go
```diff
@@ -0,0 +1,130 @@
...
+	return &SystemdCollector{
+		dbusConn: conn,
+		systemdState: typedDesc{prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "systemd", "health"),
+			"State of systemd service",
+			nil, nil,
+		), prometheus.GaugeValue},
+		systemdUnitState: typedDesc{prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "systemd", "unit_health"),
+			"State of systemd unit",
```
a-palchikov: service is also a unit

monitoring/collector/systemd.go
```diff
@@ -0,0 +1,130 @@
...
+			[]string{"unit_name"}, nil,
+		), prometheus.GaugeValue},
+	}, nil
+}
+
+func (s *SystemdCollector) Collect(ch chan<- prometheus.Metric) error {
```
a-palchikov: Should keep documenting the exports

monitoring/collector/systemd.go
```diff
@@ -0,0 +1,130 @@
...
+	systemdStatus, err := monitoring.IsSystemRunning()
+	if err != nil {
+		return trace.Wrap(err, "failed to query systemd status")
+	}
+
+	var systemdHealthMetric prometheus.Metric
+	if systemdStatus == monitoring.SystemStatusRunning {
+		if systemdHealthMetric, err = s.systemdState.newConstMetric(1.0); err != nil {
+			return trace.Wrap(err, "failed to create prometheus metric: %v", err)
```
a-palchikov: Should be enough to `trace.Wrap(err, ...)` w/o having to specify the error in the message as well.
Is there a reason to always create the metric(1.0) or metric(0.0) in `Collect` - can create the two at init and use the values here.

monitoring/collector/systemd.go
```diff
@@ -0,0 +1,130 @@
...
+		if systemdHealthMetric, err = s.systemdState.newConstMetric(0.0); err != nil {
+			return trace.Wrap(err, "failed to create prometheus metric: %v", err)
+		}
+	}
+
+	ch <- systemdHealthMetric
+
+	units, err := s.dbusConn.ListUnits()
+	if err != nil {
+		return trace.Wrap(err, "failed to query systemd units: %v", err)
```
a-palchikov: Same here about printing the error twice

monitoring/collector/systemd.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return nil
+}
+
+type loadState string
+
+const (
+	loadStateLoaded   loadState = "loaded"
+	loadStateError              = "error"
```
a-palchikov: this is no longer typed as `loadState` - you'd need to repeat the type for all constants.

monitoring/collector/systemd.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"github.com/gravitational/satellite/monitoring"
+
+	"github.com/coreos/go-systemd/dbus"
+	"github.com/gravitational/trace"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var listOfServices = map[string]int{
+	"flanneld.service":                1,
```
a-palchikov: technically, as you're not using the value, you could go with `map[string]struct{}`:
```go
var listOfServices = serviceMap(
"flanneld.service",
"serf.service",
"planet-agent.service",
)

func serviceMap(services ...string) map[string]struct{} {
  result := map[string]struct{}
  for _, service := range services {
    result[service] = struct{}{}
  }
  return result
}
```

PR #56, opened 2018-03-27
Add systemd prometheus metrics collector

UnderGreen:


monitoring/collector/systemd.go
```diff
@@ -0,0 +1,139 @@
...
+	} else {
+		ch <- systemdUnhealthy
+	}
+
+	units, err := s.dbusConn.ListUnits()
+	if err != nil {
+		return trace.Wrap(err, "failed to query systemd units")
+	}
+
+	var systemdUnitHealthMetric prometheus.Metric
```
a-palchikov: Why is this variable outside of `if`, does not seem to be required anywhere but in that `if` block

PR #57, opened 2018-03-28
Filter members in status "left" when evaluating cluster status

a-palchikov:


agent/agent.go
```diff
@@ -365,31 +376,40 @@ const statusQueryReplyTimeout = 30 * time.Second
 func (r *agent) statusUpdateLoop() {
 	statusUpdateCh := r.statusClock.After(statusUpdateTimeout)
 	recycleCh := r.recycleClock.After(recycleTimeout)
+	replyTimeout := r.statusQueryReplyTimeout
+	if replyTimeout == 0 {
+		replyTimeout = statusQueryReplyTimeout
+	}
+	// collectCh is a chan that receives updates when the status collecting process
```
r0mant: "is a channel"

agent/server_test.go
```diff
@@ -511,6 +648,14 @@ func testDialRPC(port int, certFile string) dialRPC {
 	}
 }
 
+func tags(tags ...string) map[string]string {
+	result := map[string]string{}
+	for i := 0; i < len(tags); i += 2 {
+		result[tags[i]] = tags[i+1]
```
r0mant: You'll get an "index out of bounds" panic here if somebody passes an odd number of tags.

PR #58, opened 2018-03-29
dep prune and update

knisbet:
Update to k8s 1.9 client


a-palchikov:
Hmm, I did not pay attention the first time - it looks like it vendors staging directly from kubernetes just b/c of the `SimpleNameGenerator` - I think we can just copy the couple of lines of implementation avoid pulling in far more in dependencies.

knisbet:
Abandoning in favor of #59

PR #63, opened 2018-04-10
Updated go-systemd to latest version

eldios:
Updated coreos/go-systemd to the latest version (v16).


UnderGreen:
@eldios, as I remember I set library version to 4.0 especially. Make sense to check monitoring of systemd units, maybe it is broken now.

a-palchikov:
There were no changes to `go-systemd` since migration to `dep` as far as I could tell.

eldios:
@UnderGreen thanks for mentioning that.. we were trying to understand why we were stuck with that specific version.
I tried compiling `satellite` using a simple `make` and it worked... `make test` is segfaulting for me, but I guess that's bound to the fact that I'm using Go 1.10 ?

eldios:
@a-palchikov in version 4.0.0 a few functions were not implemented yet... that's why the idea to move on to a newer version.

a-palchikov:
I meant that we did not change the version of `go-systemd` since the repository was migrated to `dep`.

eldios:
testing

eldios:
Finally manually verified, it's working. :+1:  merging.

PR #65, opened 2018-05-30
Add checker to detect whether the node has an IAM profile assigned

knisbet:
https://github.com/gravitational/telekube/issues/3453

This is to address an issue encountered at a customer site, where they accidentally lost their IAM credentials on a couple of their nodes. In our monitoring agent, if we're running on AWS, we'll poll the IAM credentials and alert if they're missing.


monitoring/aws.go
```diff
@@ -0,0 +1,69 @@
...
+
+type awsHasProfileChecker struct{}
+
+// Name returns this checker name
+// Implements health.Checker
+func (*awsHasProfileChecker) Name() string {
+	return awsHasProfileCheckerID
+}
+
+// Check will scan current open ports and report every conflict detected
```
a-palchikov: The comment has to be updated too.

knisbet:
My bad, not sure what I did to miss the Gopkg.toml

PR #66, opened 2018-06-07
Dmitri/3455 script checker

a-palchikov:
This PR adds a checker based on a bash script.

### DIFF
--- a/monitoring/script.go
+++ b/monitoring/script.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2017 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package monitoring
+
+import (
+	"context"
+	"io"
+	"io/ioutil"
+	"os/exec"
+
+	"github.com/gravitational/satellite/agent/health"
+	"github.com/gravitational/trace"
+)
+
+// NewScriptChecker returns a new script checker for the specified script
+func NewScriptChecker(script io.Reader, args ...string) health.Checker {
+	return scriptChecker{Reader: script, args: args}
+}
+
+// Name returns name of the checker.
+// Implements health.Checker
+func (r scriptChecker) Name() string {
+	return scriptCheckerID
+}
+
+func (r scriptChecker) Check(ctx context.Context, reporter health.Reporter) {
+	f, err := ioutil.TempFile("", "monitoring")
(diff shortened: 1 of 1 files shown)


monitoring/script.go
```diff
@@ -0,0 +1,79 @@
...
+
+	_, err = io.Copy(f, r.Reader)
+	if err != nil {
+		reporter.Add(NewProbeFromErr(r.Name(), "failed to write script file",
+			trace.Wrap(err)))
+		return
+	}
+
+	args := append([]string{f.Name()}, r.args...)
+	cmd := exec.Command("bash", args...)
```
knisbet: This check is run in a user's cli right, so it should always have a functional path to locate bash on the system?

knisbet: What about using CommandContext and passing the context to ensure the script times out?

monitoring/script.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/gravitational/satellite/agent/health"
+	"github.com/gravitational/trace"
+)
+
+// NewScriptChecker returns a new script checker for the specified script.
+// dir can specify an alternative working directory for the script.
+// If dir is left unspecified, the running process's directory is used
+// as a working directory.
+func NewScriptChecker(script io.Reader, dir string, args ...string) health.Checker {
```
r0mant: Call the parameter `workingDir` then?

monitoring/script.go
```diff
@@ -0,0 +1,79 @@
...
+	return scriptChecker{Reader: script, dir: dir, args: args}
+}
+
+// Name returns name of the checker.
+// Implements health.Checker
+func (r scriptChecker) Name() string {
+	return scriptCheckerID
+}
+
+func (r scriptChecker) Check(ctx context.Context, reporter health.Reporter) {
```
r0mant: Missing godoc.

monitoring/script.go
```diff
@@ -0,0 +1,79 @@
...
+	f, err := ioutil.TempFile("", "monitoring")
+	if err != nil {
+		reporter.Add(NewProbeFromErr(r.Name(), "failed to create script file",
```
r0mant: I would expand the error message, e.g. something like `failed to create script file in the temp dir %v, does it have proper permissions?` and insert the temp directory path. We've seen many cases where /tmp was not writable.

monitoring/script.go
```diff
@@ -0,0 +1,79 @@
...
+			trace.ConvertSystemError(err)))
+		return
+	}
+	defer f.Close()
+
+	_, err = io.Copy(f, r.Reader)
+	if err != nil {
+		reporter.Add(NewProbeFromErr(r.Name(), "failed to write script file",
```
r0mant: Like above: I would at least add a script file name (path) to the message.

monitoring/script.go
a-palchikov: Yes, hence I left it unspecified for `Command` to figure it out (which it will with [`LookPath`](https://golang.org/pkg/os/exec/#LookPath)) or did you mean anything else?

knisbet: No, I was just trying to make sure we don't envision any circumstances where this is run without an environment, like triggered through an rpc agent that is installed as a systemd unit.

PR #73, opened 2018-08-20
Ignoring sockets in time-wait state when checking ports.

r0mant:


monitoring/port.go
```diff
@@ -103,6 +104,14 @@ func (c *portChecker) checkProcess(proc process, reporter health.Reporter) bool
 		if r.Protocol != proc.socket.proto() {
 			continue
 		}
+		// if prechecks have been run multiple times quickly, sockets from
+		// the ping-pong test may still be lingering in time-wait state and
+		// should eventually go away
+		switch proc.socket.state() {
+		case TimeWait, Close:
```
knisbet: Why close State?

monitoring/port.go
```diff
@@ -103,6 +104,14 @@ func (c *portChecker) checkProcess(proc process, reporter health.Reporter) bool
 		if r.Protocol != proc.socket.proto() {
 			continue
 		}
+		// if prechecks have been run multiple times quickly, sockets from
+		// the ping-pong test may still be lingering in time-wait state and
+		// should eventually go away
```
knisbet: Nit: I would maybe just preface this with `ignore time-wait state`... explanation of why it's being ignored. It might be just me, but for some reason I have a hard time parsing this as is.

monitoring/port.go
r0mant: I think that it actually means `closed` state so I thought I'd add it for completeness, since it means that the socket is closed and also should be released soon.

PR #78, opened 2018-08-27
Add checker data to probes (backport 5.0).

r0mant:
Backport of https://github.com/gravitational/satellite/pull/77.


agent/proto/agentpb/agent.proto
```diff
@@ -79,20 +79,24 @@ message NodeStatus {
...
     Running = 1;
     Failed = 2;
     Terminated = 3;
   }
+  // Status is the result of the probe
   Type status = 4;
+  // Error is the probe-specific error message
   string error = 5;
+  // CheckerData is a free-form data specific to the checker
+  bytes checker_data = 6;
```
knisbet: I think it's just a style difference, but I don't think I'm a huge fan of embedding arbitrary data like this by encoding a structure to json. I realize it's only important to specific probes, but to me it seems far more straightforward to extend the message and generate compile time failures if there are breaking changes.

r0mant: Yeah, thought about it too: I didn't want to end up with the mess where this struct contains all possible fields.

PR #79, opened 2018-09-10
use mTLS for client authentication

knisbet:
Setup satellite monitoring ports to use mTLS authentication.
https://github.com/gravitational/gravity.e/issues/3726

I did a bit of refactoring to support this as well:
- When receiving an HTTP request, avoid invoking the gRPC client to localhost, and instead just invoke the gRPC service directly in process.
- I think some of the tests are broken unrelated to this change, I fixed a couple but I think some tests are still broken.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -128,14 +128,11 @@ func New(config *Config) (Agent, error) {
 		serfClient:      client,
 		name:            config.Name,
 		cache:           config.Cache,
+		dialRPC:         defaultDialRPC(config.CAFile, config.CertFile, config.KeyFile),
 		statusClock:     clock,
 		recycleClock:    clock,
 		localStatus:     emptyNodeStatus(config.Name),
 		metricsListener: metricsListener,
-		caFile:          config.CAFile,
-		certFile:        config.CertFile,
-		keyFile:         config.KeyFile,
-		rpcPort:         RPCPort,
 	}
 
 	agent.rpc, err = newRPCServer(agent, config.CAFile, config.CertFile, config.KeyFile, config.RPCAddrs)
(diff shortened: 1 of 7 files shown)


agent/client.go
```diff
@@ -44,19 +44,28 @@ type client struct {
...
-	if !certPool.AppendCertsFromPEM(cert) {
-		return nil, trace.Wrap(err, "failed to append certificates from %v", certFile)
+	if !certPool.AppendCertsFromPEM(clientCACert) {
+		return nil, trace.Wrap(err, "failed to append certificates from %v", caFile)
 	}
 
 	creds := credentials.NewTLS(&tls.Config{
-		RootCAs:    certPool,
-		MinVersion: tls.VersionTLS12,
+		RootCAs:      certPool,
```
a-palchikov: You need to be able to pass `ServerName` override here as the test certificates use `agent` as a CN.

agent/agent.go
```diff
@@ -293,7 +301,10 @@ func (r *agent) LocalStatus() *pb.NodeStatus {
 	return r.recentLocalStatus()
 }
 
-type dialRPC func(*serf.Member) (*client, error)
+func (r *agent) dialMember(member *serf.Member) (*client, error) {
+	addr := fmt.Sprintf("%s:%d", member.Addr.String(), r.rpcPort)
```
a-palchikov: This will format as `<nil>:7676` in tests (for port == `7676`) and fail to connect as the "remote" in a test is also running on localhost.

agent/agent.go
```diff
@@ -185,13 +188,12 @@ type agent struct {
 	// status sync with other agents.
 	rpc RPCServer
 
+	// rpcPort is the por this server is listening on for RPC calls
```
r0mant: port

agent/server.go
```diff
@@ -92,82 +100,61 @@ func newRPCServer(agent *agent, caFile, certFile, keyFile string, rpcAddrs []str
...
-		status, err := client.Status(context.TODO())
+		status, err := s.Status(context.TODO(), nil)
 		if err != nil {
 			roundtrip.ReplyJSON(w, http.StatusServiceUnavailable, map[string]string{"error": err.Error()})
 			return
 		}
 
 		httpStatus := http.StatusOK
-		if isDegraded(*status) {
+		if isDegraded(*status.GetStatus()) {
```
r0mant: Do you still need to dereference the pointer here?

knisbet: yes, GetStatus returns a SystemStatus pointer.

PR #80, opened 2018-09-13
Override ListenAddr in port checker to be able to test only specific IP addresses

a-palchikov:
- Adds `ListenAddr` override to be able to restrict open port checker to the given local address. 
 - Adds missing tests (also one from kevin's PR).

### DIFF
--- a/monitoring/port.go
+++ b/monitoring/port.go
@@ -109,18 +109,13 @@ func (c *portChecker) checkProcess(proc process, reporter health.Reporter) bool 
 		}
 		switch proc.socket.state() {
 		case TimeWait:
-			// ignore sockets in time-wait and closed states since they're going
+			// ignore sockets in time-wait state since they're going
 			// away soon
 			log.Debugf("Ignoring %v for program %q(pid=%v).", formatSocket(proc.socket), proc.name, proc.pid)
 			continue
 		}
-		localAddr := proc.localAddr().ip.String()
-		addr := r.ListenAddr
-		if addr == "" {
-			addr = localAddr
-		}
-		if uint64(proc.localAddr().port) >= r.From && uint64(proc.localAddr().port) <= r.To &&
-			(proc.localAddr().ip.IsUnspecified() || localAddr == addr) {
+
+		if isPortConflict(proc, r) && isIPConflict(proc, r) {
 			conflicts = true
 			reporter.Add(&pb.Probe{
 				Checker: portCheckerID,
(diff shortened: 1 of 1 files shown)


monitoring/port.go
```diff
@@ -104,14 +107,20 @@ func (c *portChecker) checkProcess(proc process, reporter health.Reporter) bool
...
 			continue
 		}
-		if uint64(proc.localAddr().port) >= r.From && uint64(proc.localAddr().port) <= r.To {
+		localAddr := proc.localAddr().ip.String()
+		addr := r.ListenAddr
+		if addr == "" {
+			addr = localAddr
+		}
+		if uint64(proc.localAddr().port) >= r.From && uint64(proc.localAddr().port) <= r.To &&
+			localAddr == addr {
```
knisbet: I think this will fail to detect services bound to any address, ie 0.0.0.0, which I believe will conflict with a bind to a specific address.

monitoring/port_test.go
```diff
@@ -146,6 +146,52 @@ func (*MonitoringSuite) TestValidatesPorts(c *C) {
...
+				},
+			),
+			probes: []*pb.Probe{
+				&pb.Probe{
+					Checker: portCheckerID,
+					Status:  pb.Probe_Running,
+				},
+			},
+			comment: "ignores open ports in TimeWait",
+		},
```
knisbet: Here's a test case that I think will detect the ListenAddr set, but an any bind:
```
		{
			ports: []PortRange{
				PortRange{From: 9001, To: 9001, Protocol: "tcp", ListenAddr: "127.0.0.2"},
			},
			reader: testGetProcs(
				process{
					name:   "foo",
					pid:    100,
					socket: tcpSocket{State: Listen, LocalAddress: tcpAddr("0.0.0.0:9001", c)},
				},
			),
			probes: []*pb.Probe{
				&pb.Probe{
					Checker: portCheckerID,
					Detail:  `conflicting program "foo"(pid=100) is occupying port tcp/9001(listen)`,
					Status:  pb.Probe_Failed,
				},
			},
			comment: "detects process bound to any address",
		},
```

monitoring/port.go
a-palchikov: Good catch, Kevin

monitoring/port_test.go
a-palchikov: :heavy_check_mark:

monitoring/port.go
```diff
@@ -104,14 +107,20 @@ func (c *portChecker) checkProcess(proc process, reporter health.Reporter) bool
...
 			log.Debugf("Ignoring %v for program %q(pid=%v).", formatSocket(proc.socket), proc.name, proc.pid)
 			continue
 		}
-		if uint64(proc.localAddr().port) >= r.From && uint64(proc.localAddr().port) <= r.To {
+		localAddr := proc.localAddr().ip.String()
+		addr := r.ListenAddr
+		if addr == "" {
+			addr = localAddr
+		}
+		if uint64(proc.localAddr().port) >= r.From && uint64(proc.localAddr().port) <= r.To &&
```
r0mant: This is more of a nitpick, but I think the readability of this condition would greatly improve if you made a couple of helper functions like:

```go
func isPortConflict(proc <type-of-proc>, range <type-of-r>) bool {}
func isIPConflict(proc <type-of-proc>, range <type-of-r>) bool {}

if isPortConflict(proc, r) && isIPConflict(proc, r) {}
```

And the code above that determines addr could also move to isIPConflict function.

PR #81, opened 2018-09-14
Add a checker that tests if inotify is able to create watches

knisbet:
I've seen a couple customers run into a scenario where they've run out of inotify watches. Unfortunately, the kernel doesn't have a metric that makes this easy to poll and set a threshold, but we can do a simple test on whether we're able to create a watch or not. This is the same principal used in the prometheus node exporter to detect this type of issue.

Also, I'm planning to add a sysctl check, but will do that as a separate PR.

### DIFF
--- a/monitoring/defaults_linux_test.go
+++ b/monitoring/defaults_linux_test.go
@@ -1,0 +1,37 @@
+/*
+Copyright 2016 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package monitoring
+
+import (
+	"context"
+
+	"github.com/gravitational/satellite/agent/health"
+	"gopkg.in/check.v1"
+)
+
+type InotifySuite struct{}
+
+func (_ *InotifySuite) TestCreateInotify(c *check.C) {
+	// Note, it's difficult to simulate on linux
+	// running out of inotify watches, so we only
+	// test the success case here.
+
+	checker := &iNotifyChecker{}
+	var probes health.Probes
+	checker.Check(context.TODO(), &probes)
+	c.Assert(probes.NumProbes(), check.Equals, 0)
+}
--- a/monitoring/inotify.go
+++ b/monitoring/inotify.go
@@ -1,0 +1,27 @@
+// +build !linux
+
+/*
+Copyright 2017 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package monitoring
+
+import (
+	"github.com/gravitational/satellite/agent/health"
+)
+
+func NewINotifyChecker() health.Checker {
+	return noopChecker{}
+}
--- a/monitoring/inotify.go
+++ b/monitoring/inotify.go
@@ -22,9 +22,9 @@ import (
 	"github.com/gravitational/satellite/agent/health"
 )
 
-// NewINotifyChecker creates a new health.Checker that tests if inotify watches
+// NewInotifyChecker creates a new health.Checker that tests if inotify watches
 // can be created. This is usually an indication that the system has reached
 // fs.inotify.max_user_instances has been exhausted
-func NewINotifyChecker() health.Checker {
+func NewInotifyChecker() health.Checker {
 	return noopChecker{}
 }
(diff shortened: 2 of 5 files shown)


monitoring/defaults_linux_test.go
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright 2016 Gravitational, Inc.
```
r0mant: 2018

monitoring/inotify.go
```diff
@@ -0,0 +1,27 @@
...
+limitations under the License.
+*/
+
+package monitoring
+
+import (
+	"github.com/gravitational/satellite/agent/health"
+)
+
+func NewINotifyChecker() health.Checker {
```
r0mant: Godoc plz.

monitoring/inotify_linux.go
```diff
@@ -0,0 +1,59 @@
+/*
+Copyright 2017 Gravitational, Inc.
```
r0mant: 2018

monitoring/inotify_linux.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+
+	"github.com/gravitational/satellite/agent/health"
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+
+	"golang.org/x/sys/unix"
+)
+
+func NewINotifyChecker() health.Checker {
```
r0mant: Godoc.

monitoring/inotify_linux.go
```diff
@@ -0,0 +1,59 @@
...
+// iNotifyChecker tries to create an inotify watch, and reports if it fails
+type iNotifyChecker struct {
+}
+
+// Name returns name of the checker
+func (c iNotifyChecker) Name() string {
+	return "inotify"
+}
+
+func (c iNotifyChecker) Check(ctx context.Context, reporter health.Reporter) {
```
r0mant: Godoc.

monitoring/inotify_linux.go
```diff
@@ -0,0 +1,59 @@
...
+	fd, err := unix.InotifyInit()
```
r0mant: Error is not checked.

monitoring/inotify_linux.go
```diff
@@ -0,0 +1,59 @@
...
+	if fd < 0 {
+		reporter.Add(&pb.Probe{
+			Checker: c.Name(),
+			Status:  pb.Probe_Failed,
+			Error:   "Unable to initialize inotify",
+		})
```
r0mant: Should there be a return here?

monitoring/inotify_linux.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	defer unix.Close(fd)
+
+	if _, err = unix.InotifyAddWatch(fd, "/", unix.IN_CREATE); err != nil {
+		reporter.Add(&pb.Probe{
+			Checker: c.Name(),
+			Status:  pb.Probe_Failed,
+			Error:   "Unable to create inotify watch",
```
r0mant: I think normally the Error field holds the error message (`trace.UserMessage(err)`) and user-friendly description goes into Detail. Otherwise the original error message is lost. Same for the failed probe above.

monitoring/inotify_linux.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// Name returns name of the checker
+func (c iNotifyChecker) Name() string {
+	return "inotify"
+}
+
+// Check tests whether inotify is working
+func (c iNotifyChecker) Check(ctx context.Context, reporter health.Reporter) {
+	fd, err := unix.InotifyInit()
```
r0mant: Can this return non-nil error and fd >= 0 (e.g. fd == 0)? If so, this case will be missed. I would modify the condition below to `if fd <0 || err != nil`, then all error cases should be covered.

knisbet: I don't believe this is possible, this is basically a direct mapping to the linux syscall, which used an FD of -1 to indicate error.

monitoring/inotify_linux_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+
+	"github.com/gravitational/satellite/agent/health"
+	"gopkg.in/check.v1"
+)
+
+type InotifySuite struct{}
+
+func (_ *InotifySuite) TestCreateInotify(c *check.C) {
+	// Note, it's difficult to simulate on linux
```
a-palchikov: Not if you abstract `InotifyInit` and `InotifyAddWatch` into `Init` and `AddWatch` and simply simulate whatever conditions you're interested in.

monitoring/inotify_linux.go
```diff
@@ -0,0 +1,67 @@
...
+			Checker: c.Name(),
+			Status:  pb.Probe_Failed,
+			Detail:  "Unable to initialize inotify",
+			Error:   trace.UserMessage(trace.ConvertSystemError(err)),
+		})
+		return
+	}
+	defer unix.Close(fd)
+
+	if _, err = unix.InotifyAddWatch(fd, "/", unix.IN_CREATE); err != nil {
```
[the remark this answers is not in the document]
knisbet: InotifyAddWatch just returns a watch descriptor, basically the table entry in the watch instance used for matching events. This isn't exploiting behaviour at all, it's just useless information in the context of whether we're able to add a watch to an inotify instance. It's also unnecessary to remove, because the watch table will get removed when the inotify fd is closed with the defer in line 57.

And unless I'm missing something, monitoring `/` is a file that's guaranteed to exist and that every user has read access to (which is the requirement to add a watch). Otherwise, we either need to pass configuration, or make assumptions about the environment and user the checker is running as. I'm not sure why this is overkill in permissions, to me it seems like the method that requires the least permissions, configuration, and assumptions.

monitoring/inotify_linux_test.go
a-palchikov: Not sure what you mean by 18 line function becoming completely unreadable with the use of abstractions instead of real syscalls. If you had it, you could mock whatever results you needed - you're not using that much of the functionality to not be able to do that.

PR #83, opened 2018-09-27
Dmitri/3180 node status

a-palchikov:
This PR adds a new checker to verify the availability of a single kubernetes node.
Note, that we already have a checker to validate similar facts but it works as part of a parallel functionality that provides information on all the nodes but also has a configurable failure threshold.

There's also new probe type (`temporary`) to denote a temporary failure which does not make the `status` query fail. Only the `HTTPS` endpoint will report `503` for temporary errors (currently only `NodeNotReady` for individual nodes).

As a side-effect, I had to migrate kubernetes access configuration code to use clients unequivocally and universally and upgraded versions of `protoc` and `Go` (in `buildbox` image) to match the main user of this library.


monitoring/nodesstatus.go
```diff
@@ -94,3 +88,87 @@ func (r *nodesStatusChecker) Check(ctx context.Context, reporter health.Reporter
...
 	}
 }
+
+// NewNodeStatusChecker returns a Checker that validates availability
+// of a single kubernetes node
+func NewNodeStatusChecker(config KubeConfig, nodeName string) health.Checker {
+	nodeLister := kubeNodeLister{client: config.Client.CoreV1()}
+	return &nodeStatusChecker{
+		nodeLister: nodeLister,
+		nodeName:   nodeName,
```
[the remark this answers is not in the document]
a-palchikov: It should be the node name from the cloud provider (the one we query from metadata) if that's the case. It is also the name that comes into planet as `KUBE_NODE_NAME` envar.

monitoring/nodesstatus.go
```diff
@@ -94,3 +88,87 @@ func (r *nodesStatusChecker) Check(ctx context.Context, reporter health.Reporter
...
+}
+
+type nodeLister interface {
+	Nodes() (*v1.NodeList, error)
+}
+
+func (r kubeNodeLister) Nodes() (*v1.NodeList, error) {
+	options := metav1.ListOptions{
+		LabelSelector: labels.Everything().String(),
+		FieldSelector: fields.Everything().String(),
```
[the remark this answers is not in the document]
a-palchikov: Oh, that's a good idea - haven't thought of that, just copied the code from the other node status checker.

monitoring/nodesstatus.go
```diff
@@ -94,3 +88,87 @@ func (r *nodesStatusChecker) Check(ctx context.Context, reporter health.Reporter
...
+	if err != nil {
+		reporter.Add(NewProbeFromErr(r.Name(), trace.UserMessage(err), trace.Wrap(err)))
+		return
+	}
+
+	var unavailableNode *v1.Node
+L:
+	for _, node := range nodes.Items {
+		for _, condition := range node.Status.Conditions {
+			if condition.Type != v1.NodeReady {
```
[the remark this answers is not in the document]
a-palchikov: see below

cmd/agent/main.go
```diff
@@ -50,7 +50,7 @@ func run() error {
 		// `agent` command
 		cagent                      = app.Command("agent", "Start monitoring agent")
 		cagentRPCAddrs              = ListFlag(cagent.Flag("rpc-addr", "List of addresses to bind the RPC listener to (host:port), comma-separated").Default("127.0.0.1:7575"))
-		cagentKubeAddr              = cagent.Flag("kube-addr", "Address of the kubernetes API server").Default("http://127.0.0.1:8080").String()
+		cagentKubeconfig            = cagent.Flag("kubeconfig", "Absolute path to the kubeconfig file").OverrideDefaultFromEnvar("KUBECONFIG").Required().String()
```
r0mant: Move KUBECONFIG env var name to constants?

monitoring/nodesstatus.go
```diff
@@ -94,3 +88,87 @@ func (r *nodesStatusChecker) Check(ctx context.Context, reporter health.Reporter
...
+func (r *nodeStatusChecker) Check(ctx context.Context, reporter health.Reporter) {
+	nodes, err := r.nodeLister.Nodes()
+	if err != nil {
+		reporter.Add(NewProbeFromErr(r.Name(), trace.UserMessage(err), trace.Wrap(err)))
+		return
+	}
+
+	var unavailableNode *v1.Node
+L:
+	for _, node := range nodes.Items {
```
r0mant: I'm curious why you decided to iterate over all nodes to find the local one instead of just querying it via filter parameter.

agent/status.go
```diff
@@ -76,11 +76,20 @@ func nodeToSystemStatus(status pb.NodeStatus_Type) pb.SystemStatus_Type {
...
+	status := resp.GetStatus()
 	switch status.Status {
 	case pb.SystemStatus_Unknown, pb.SystemStatus_Degraded:
 		return true
 	}
+	for _, node := range status.Nodes {
+		for _, probe := range node.Probes {
+			if probe.Status == pb.Probe_Temporary {
+				// Consider temporary failures as degrading the cluster state
+				return true
```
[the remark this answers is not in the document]
a-palchikov: Because as I wrote in description, only the `healthz` endpoint is sensitive to warnings - otherwise, only critical errors cause the system status to go `degraded`. it's something we talked about with @klizhentas.

monitoring/nodesstatus.go
a-palchikov: This has been changed to use the filter.

agent/status.go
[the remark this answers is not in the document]
a-palchikov: I reverted these changes as they seem to break symmetry of `status` and healthz results.

agent/health/health.go
```diff
@@ -67,6 +67,9 @@ type Probes []*pb.Probe
 // Add adds a health probe for a specific node.
 // Implements Reporter
 func (r *Probes) Add(probe *pb.Probe) {
+	if probe.Status == pb.Probe_Failed && probe.Severity == pb.Probe_Cleared {
```
r0mant: Not sure I understand this logic. Could you explain why severity is being adjusted here, maybe add a comment?

Is this to set severity on probes that do not set it themselves?

agent/proto/agentpb/agent.proto
```diff
@@ -92,12 +92,24 @@ message Probe {
     Failed = 2;
     Terminated = 3;
   }
+  // Severity defines the severity of the probe.
+  enum Severity {
+    // Cleared severity denotes the severity of a running probe
+    Cleared = 0;
```
r0mant: It's not very clear what "cleared" means. No pun intended :) Does it effectively mean "success"?

agent/health/health.go
a-palchikov: It was meant as a refactoring aid to match the old behavior - whenever a failed probe is added w/o an explicitly set severity, it is reset to `Critical`.

PR #84, opened 2018-10-03
coredns

knisbet:
Note: Do not merge until 5.2 maintenance branch created

Add a check to test DNS resolution for CoreDNS by querying the cluster leader. Unfortunately, it's not currently possible to also query the k8s service discovery, due to a dependency cycle on the checks passing in order to complete provisioning coredns rbac permissions, so just test leader election.

Updates https://github.com/gravitational/gravity.e/issues/3742


monitoring/defaults_linux.go
```diff
@@ -128,3 +128,12 @@ func DefaultBootConfigParams() health.Checker {
 		},
 	)
 }
+
+// NewDNSChecker sends some default queries to monitor DNS / service discovery health
+func NewDNSChecker() health.Checker {
+	return &DNSChecker{
+		QuestionA: []string{
+			"leader.telekube.local",
```
r0mant: I think this should be a parameter to the `NewDNSChecker` function, satellite should not know anything about our internal DNS names (planet knows it and can pass it here).

PR #85, opened 2018-10-06
Fix Mac build.

r0mant:


monitoring/defaults.go
```diff
@@ -83,4 +88,33 @@ type StorageConfig struct {
 	Filesystems []string
 	// MinFreeBytes define minimum free volume capacity
 	MinFreeBytes uint64
+	// HighWatermark is the disk occupancy percentage that is considered degrading
```
a-palchikov: I think it makes sense to move `StorageConfig` and `HighWatermarkCheckerData` to monitoring/storage.go (not restricted to any particular arch/OS) so it's not duplicated. monitoring/defaults.go is not the proper place for this.

PR #87, opened 2018-11-28
Handle a '/local' path with local node status results.

a-palchikov:
Updates https://github.com/gravitational/gravity.e/issues/3830.

### DIFF
--- a/agent/server_test.go
+++ b/agent/server_test.go
@@ -385,7 +385,7 @@ func (r *AgentSuite) TestIsMember(c *C) {
 	c.Assert(agent.IsMember(), Equals, true)
 }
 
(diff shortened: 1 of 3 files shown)


agent/server_test.go
```diff
@@ -704,7 +749,7 @@ type byChecker []*pb.Probe
 type tags map[string]string
 
 func generateCert(certFile, keyFile string) error {
-	priv, err := ecdsa.GenerateKey(elliptic.P224(), rand.Reader)
+	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
a-palchikov: This curve seems to be compatible with `TLS1.3`

PR #89, opened 2018-12-07
Detect client closed situations and reinit the client.

a-palchikov:
This is to fix the issues with serf dropping the connection (i.e. as a result of restarting the serf unit).

### DIFF
--- a/agent/serf.go
+++ b/agent/serf.go
@@ -17,6 +17,8 @@ limitations under the License.
 package agent
 
 import (
+	"sync"
+
 	"github.com/gravitational/trace"
 	serf "github.com/hashicorp/serf/client"
 )
@@ -54,6 +56,8 @@ func (r *retryingClient) Members() ([]serf.Member, error) {
 	if err := r.reinit(); err != nil {
 		return nil, trace.Wrap(err)
 	}
+	r.RLock()
+	defer r.RUnlock()
 	return r.client.Members()
 }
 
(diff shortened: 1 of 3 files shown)


agent/serf.go
```diff
@@ -25,15 +26,92 @@ import (
...
+
+// Close closes the client
+func (r *retryingClient) Close() error {
+	if r.client.IsClosed() {
+		return nil
+	}
+	return r.client.Close()
+}
+
+func (r *retryingClient) reinit() error {
```
r0mant: Can this client be shared by multiple goroutines? If so, this method (and Close too) should be guarded by a mutex.

agent/serf.go
```diff
@@ -25,15 +26,92 @@ import (
...
+	if !r.client.IsClosed() {
```
knisbet: Looking at https://github.com/hashicorp/serf/blob/master/client/rpc_client.go It looks to me like IsClosed is only set when RPCClient.Close is called, unless I'm missing it I don't see another path.

Are you sure IsClosed is set when we're encountering this issue? To me it looks like we should probably be receiving some sort of IO error, and may need to try and reconnect if io errors are encountered.

a-palchikov: Client will be closed whenever listen fails.
https://github.com/hashicorp/serf/blob/v0.8.1/client/rpc_client.go#L147:
```go
func ClientFromConfig(c *Config) (*RPCClient, error) {
	// ...
	go client.listen()
	// ...
}
func (c *RPCClient) listen() {
	defer c.Close()
	// ...
}
```

knisbet: Ok thanks... just wanted to be sure as on initial inspection it didn't look like it did :)

PR #96, opened 2019-02-12
Add pingCheck to check ping time between master nodes

eldios:
## Task

Add pingCheck to check ping time between master nodes.

### Brief
In order to avoid false positives and noise, a sliding window of x (10 by default) pings should be used to calculate the actual final value to compare to the threshold.
Golang `hdrhistogram` will be used to calculate if the y (95 by default) percentile is larger than the threshold.

* If nodeType != Master
    * NOOP

* if nodeType == Master
    * find other masters
    * ping them
    * store time via `serf`
    * If pingTime >= threshold then send alert.

### Complete 
[Google Drive Spec Doc](https://docs.google.com/document/d/1XNqdW1m18tJ3vAYAPdFnZnU_GI1WST33SQWIwZVzY0A/edit#heading=h.cvo0tmjkja8c)


eldios:
not finished yet, asking for comments mostly to verify I'm going in the right direction and continue with the wrap up.

cmd/agent/main.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/gravitational/version"
 
 	log "github.com/sirupsen/logrus"
-	"gopkg.in/alecthomas/kingpin.v2"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
```
a-palchikov: This is no-op change - the package is already called `kingpin`

cmd/agent/checkers.go
```diff
@@ -68,6 +68,7 @@ func addToMaster(node agent.Agent, config *config, kubeConfig monitoring.KubeCon
 	node.AddChecker(monitoring.DockerHealth(config.dockerAddr))
 	node.AddChecker(etcdChecker)
 	node.AddChecker(monitoring.SystemdHealth())
+	node.AddChecker(monitoring.PingHealth(config.serfRPCAddr,config.role)
```
a-palchikov: nit: missing white-space before the comma - `go fmt` should've fixed that.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+func NewPingChecker(serfRPCaddr string, role string) health.Checker {
+	return &PingChecker{
+		serfRPCAddr: serfRPCaddr,
+		role:        role,
+	}
+}
+
+// PingChecker is a checker that verifies if ping between Master Nodes and other
```
a-palchikov: nit: I would not capitalize `Master` (and `Nodes`) - nothing special about them.
The comment could be more clear (also threshold is missing b/c it's supposed to be `specified`):
```go
// PingChecker is a checker to verify that ping times between nodes in the cluster are within a predefined
// threshold.
```
Note that mentioning master (vs other ones) is leaking implementation detail and I don't think is necessary.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/gravitational/satellite/agent"
+	"github.com/gravitational/satellite/agent/health"
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+	"github.com/gravitational/trace"
+	serf "github.com/hashicorp/serf/client"
+	ping "github.com/sparrc/go-ping"
+)
+
+const (
+	PingCheckerID = "ping-checker"
```
a-palchikov: this could be package-local if not used outside of the package.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// PingChecker is a checker that verifies if ping between Master Nodes and other
+// ones is below a specified threshold
+type PingChecker struct {
+	serfRPCAddr string
+	role        string
+}
+
+// Name returns the check name
```
a-palchikov: check**er** name

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+// Implements health.Checker
+func (c *PingChecker) Name() string {
+	return PingCheckerID
+}
+
+// Check verifies that all nodes' ping with Master Nodes is lower than the
+// desired threshold
+// Implements health.Checker
+func (c *PingChecker) Check(ctx context.Context, r health.Reporter) {
+	RttThreshold := 100     // ms
```
a-palchikov:
```go
const rttThreshold = 100 * time.Millisecond  // self-explanatory
const slidingWindowSize = 10
```

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+	slidingWindowSize := 10 // number of ping results to consider per iteration
+
+	// set Probe to be running
+	r.Add(&pb.Probe{
+		Checker: c.Name(),
+		Status:  pb.Probe_Running,
+	})
+
+	// fetch serf config and intantiate client
```
a-palchikov: You don't really need to comment each step unless it's unclear from context.
intantiate -> instantiate

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+	clientConfig := serf.Config{
+		Addr: c.serfRPCAddr,
+	}
+	client, err := serf.ClientFromConfig(clientConfig)
+	if err != nil {
+		return // nil, trace.Wrap(err, "failed to connect to serf")
```
a-palchikov: Why is this not an error?
The way other checker handle this is by marking the resulting probe as failed.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	defer client.Close()
+
+	// only run PingCheck between masters
+	role := c.role
```
a-palchikov: you don't need a local variable - `c.role` is not that long

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package monitoring
+
+import (
+	"context"
+	"log"
```
a-palchikov: We don't use log from the standard library but instead use [this package](github.com/sirupsen/logrus)

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Check verifies that all nodes' ping with Master Nodes is lower than the
+// desired threshold
+// Implements health.Checker
+func (c *PingChecker) Check(ctx context.Context, r health.Reporter) {
+	RttThreshold := 100     // ms
+	slidingWindowSize := 10 // number of ping results to consider per iteration
+
+	// set Probe to be running
+	r.Add(&pb.Probe{
```
a-palchikov: The result of Check should be a single probe that describes the result of the validation - it is either running or failed.
This creates a running probe w/o validating anything.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+		r.Add(&pb.Probe{
+			Checker: c.Name(),
+			Status:  pb.Probe_Terminated,
+		})
+		return // skip check if the node is not a master
+	}
+
+	// ping other Master nodes and store results in Serf
+	nodes := client.Members()
+	for node := range nodes {
```
a-palchikov:
```go
for node := range client.Members() {
```

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+		// FIXME: BEGINIF if other node is master {
+		pinger, err := ping.NewPinger(node.Addr)
+		if err != nill {
+			log.Printf("got an error while trying to ping %v - %v", node.Addr,
```
a-palchikov: This should result in a failed probe

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	"context"
+	"log"
+
+	"github.com/gravitational/satellite/agent"
+	"github.com/gravitational/satellite/agent/health"
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+	"github.com/gravitational/trace"
+	serf "github.com/hashicorp/serf/client"
+	ping "github.com/sparrc/go-ping"
```
a-palchikov: this dependency seems to be missing from `Gopkg.toml`

monitoring/serf.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"sync"
+
+	"github.com/gravitational/trace"
+	serf "github.com/hashicorp/serf/client"
+)
+
+// serfClient is the minimal interface to the serf cluster.
+// It enables mocking access to serf network in tests.
+type serfClient interface {
```
a-palchikov: Why do you need this? I would export the bits from the agent package so these could be used if necessary.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+				trace.Wrap(err))
+			return
+		}
+		pinger.Count = slidingWindowSize // FIXME: does need to be set to actually use the last nth check results?
+		pinger.Run()
+		stats := pinger.Statistics()
+		if stats.AvgRtt <= RttThreshold {
+			r.Add(&pb.Probe{
+				Checker: c.Name(),
+				Status:  pb.Probe_Critical,
```
a-palchikov: This is probe's Severity, not type.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+			})
+		}
+
+		r.Add(&pb.Probe{
+			Checker: c.Name(),
+			Status:  pb.Probe_Terminated,
```
a-palchikov: I don't think we use probe types beyond `Running` and `Failed` so there maybe inconsistencies as to how terminated is regarded (whether it is a positive outcome).

cmd/agent/main.go
eldios: it was VSCode doing that through `auto-GoFmt` :man_shrugging: I'll revert :+1:

cmd/agent/checkers.go
r0mant: Looks like the closing `)` is missing. Does this compile at all?

eldios: fixed and activated `gofmt` onSave (on my current editor)

eldios: yeah, fixed... not sure how that got to be missing.

monitoring/serf.go
r0mant: Yeah, isn't this file basically a verbatim copy of `agent/serf.go`? Why not import it from there?

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+		})
+		return // skip check if the node is not a master
+	}
+
+	// ping other Master nodes and store results in Serf
+	nodes := client.Members()
+	for node := range nodes {
+		// FIXME: BEGINIF if other node is master {
+		pinger, err := ping.NewPinger(node.Addr)
+		if err != nill {
```
r0mant: `nill`- > `nil`. Please make sure the code compiles before sending PRs :)

monitoring/ping.go
r0mant: I would make threshold a part of the checker configuration. Also 100ms seems too lenient, I'd consider everything above 15-20ms (maybe even 10ms) latency on an internal network a degraded performance.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+func (c *PingChecker) Name() string {
+	return PingCheckerID
+}
+
+// Check verifies that all nodes' ping with Master Nodes is lower than the
+// desired threshold
+// Implements health.Checker
+func (c *PingChecker) Check(ctx context.Context, r health.Reporter) {
+	RttThreshold := 100     // ms
+	slidingWindowSize := 10 // number of ping results to consider per iteration
```
r0mant: I'd move this to a constant (out of the method).

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+	clientConfig := serf.Config{
+		Addr: c.serfRPCAddr,
+	}
+	client, err := serf.ClientFromConfig(clientConfig)
+	if err != nil {
+		return // nil, trace.Wrap(err, "failed to connect to serf")
+	}
+	defer client.Close()
+
+	// only run PingCheck between masters
```
r0mant: I think the comment is a bit poorly worded - it gives an impression that only master nodes ping each other, while in fact each master pings all other nodes.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+	role := c.role
+	if role != agent.RoleMaster {
```
r0mant: Instead of checking this internally, why just not make sure that this probe is added only on masters? I.e. do this on the caller side - where checkers are initialized. Otherwise this probe will be added to all regular nodes and will be firing up without actually doing anything.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+			Status:  pb.Probe_Terminated,
+		})
+		return // skip check if the node is not a master
+	}
+
+	// ping other Master nodes and store results in Serf
+	nodes := client.Members()
+	for node := range nodes {
+		// FIXME: BEGINIF if other node is master {
+		pinger, err := ping.NewPinger(node.Addr)
```
r0mant: This library sends ICMP packets as far as I can see, just like regular ping. Are you sure we can rely on ICMP not being blocked b/w the nodes? I think this is often the case.

monitoring/ping.go
```diff
@@ -0,0 +1,116 @@
...
+		if err != nill {
+			log.Printf("got an error while trying to ping %v - %v", node.Addr,
+				trace.Wrap(err))
+			return
+		}
+		pinger.Count = slidingWindowSize // FIXME: does need to be set to actually use the last nth check results?
+		pinger.Run()
+		stats := pinger.Statistics()
+		if stats.AvgRtt <= RttThreshold {
```
r0mant: We should be using https://godoc.org/github.com/codahale/hdrhistogram and percentiles, not average.

monitoring/ping.go
eldios: Comment changed.

monitoring/ping.go
eldios: Added.

monitoring/ping.go
eldios: Changed.

monitoring/ping.go
eldios: okay I guess I'll just get rid of this then?

monitoring/ping.go
eldios: ouch, sorry about that.

monitoring/ping.go
eldios: I'm not sure I got your comment @a-palchikov ...

The PingCheckedID is returned by the Name() function so I presume it's used somewhere else as all Add.Probe() function use it as well.

monitoring/serf.go
eldios: I was thinking about that but in order to do that, shouldn't I change `serfClient` to be `SerfClient` thus needing a refactoring in all the `agent` package where that's used?

Is that okay?

eldios: same answer, I gave above, but:
I was thinking about that but in order to do that, shouldn't I change `serfClient` to be `SerfClient` thus needing a refactoring in all the `agent` package where that's used?

Is that okay?

monitoring/ping.go
eldios: okay, that's definitely one of the feedback I wanted/needed. I was thinking of going that direction but didn't know where to go.
If that's okay I'll start with just using a more sensible (but conservative value) like 25ms and then work on making it a parameter later.

monitoring/ping.go
eldios: moved. :+1:

monitoring/ping.go
eldios: okay so I got it wrong.. I actually thought that we only wanted master nodes to ping other master nodes. I didn't know how do to it and thus I put the `FIXME` comment part for later, but I actually meant to create a check that would verify ping only between masters.

If we want masters nodes to ping all others (which made more sense to me) it's easier and I can definitely do it now (and remove the FIXME part).

monitoring/ping.go
eldios: Good point.
I am actually already doing this by only calling it in the `cmd/agent/checkers.go:addToMasters()` function ... so I guess that's just redundant.

monitoring/ping.go
eldios: Yes, that's a normal ICMP ping so I guess that could be true in some ENVs.

Well, the two other alternatives I evaluated were:
1- create a listener on a custom port and send/receive messages on that instead, but we'll have to ask for an additional port which is even worse probably, as opposed to allow ICMP internally
2- I guess it could be possible to send/receive messages from within the checker itself sending a fixed size packet but that would essentially mean reimplementing a "simpler ping"... right?

monitoring/serf.go
r0mant: Sure, I see nothing wrong with exporting serf client. Moreover, I bet your refactoring will be very short - just capitalize "s" in a couple of spots :)

monitoring/ping.go
r0mant: Sounds good. I would also make sure we're logging these metrics so we can for example look at the collected data and adjust the value appropriately if needed.

monitoring/ping.go
r0mant: You don't necessarily need an additional port, agent already serves API on port 7575 so you could just use that.

Also, I wonder if serf itself already has some built-in mechanism to determining latency b/w its nodes? I would assume it does, with it being a cluster membership tool. I suggest you research it, e.g. take a look at https://www.serf.io/docs/commands/rtt.html and see if we can use it instead.

eldios: Seems that Serf RTT is a really crude implementation done in CLI and returning strings

https://github.com/hashicorp/serf/blob/master/cmd/serf/command/rtt.go

doesn't seem like a good choice to me :man_shrugging:

monitoring/ping.go
eldios: okay, value changed and FIXME added.

monitoring/serf.go
eldios: I didn't mind changing variables in my code but I was more concerned about changing it everywhere in `satellite`... if that's okay I can definitely do it. Just didn't want to assume it was a good thing without @a-palchikov and @r0mant approval.
I'll go ahead and do it then.

monitoring/ping.go
r0mant: Well, I wasn't proposing to call a CLI command obviously. It uses serf API, doesn't it?

eldios: it's actually just creating an RPCClient and then use:

https://github.com/hashicorp/serf/blob/master/client/rpc_client.go#L370

on node1 and node2, then calling:

https://github.com/hashicorp/serf/blob/master/coordinate/coordinate.go#L119

between the two. I guess I could use the same approach and call those APIs myself, just not the RTT tool directly.

monitoring/serf.go
eldios: done.

monitoring/ping.go
eldios: Removed.

monitoring/serf.go
eldios: Done too.

monitoring/ping.go
eldios: think I got it... and fixed :)

monitoring/ping.go
eldios: added ... but we might change it and remove it after all.

monitoring/ping.go
eldios: fixed.

monitoring/ping.go
eldios: changed... not sure if in the right way. Feel free to reopen a different comment on the specific section if I did it wrong.

monitoring/ping.go
eldios: I tried and failed implementing this, so for now I guess we'll have all the rest set and then I'll wrap it up by adding this one last piece.

r0mant: @eldios What's the issue? Serf "rtt" command that you linked to the source of above does exactly what we need, no? So you can just implement our check same way?

eldios: Managed to implement it.

eldios:
> Also, this PR implements the latency check but not the time skew check. Are you going to add it later / in a separate PR?

Confirmed, that will be part of a separated PR/work/task.

eldios:
For completeness, this check will be tested manually using something like:

```
tc qdisc add dev eth0 root netem delay 100ms
```

eldios:
Well now the check itself is working .... except that it's not. Trying adding some latency with `tc` as depicted above didn't change the results:
```
Feb 27 01:51:20 stolonnode-1 planet[3319]: time="2019-02-27T01:51:20Z" level=debug msg="Running checker \"ping-checker\"."                                                                                                                          
Feb 27 01:51:20 stolonnode-1 planet[3319]: time="2019-02-27T01:51:20Z" level=debug msg="172_28_128_131.gra <-ping-> 172_28_128_132.gra = 0"                                                                                                         
Feb 27 01:51:20 stolonnode-1 planet[3319]: time="2019-02-27T01:51:20Z" level=debug msg="ping value 0 below threshold 15 ms"                                                                                                                         
Feb 27 01:51:20 stolonnode-1 planet[3319]: time="2019-02-27T01:51:20Z" level=debug msg="172_28_128_131.gra <-ping-> 172_28_128_133.gra = 0"                                                                                                         
Feb 27 01:51:20 stolonnode-1 planet[3319]: time="2019-02-27T01:51:20Z" level=debug msg="ping value 0 below threshold 15 ms"            
```
I'll try to see if I did anything wrong with the Serf Ping/RTT implementation.

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+			r.Add(&pb.Probe{
+				Checker: c.Name(),
+				Status:  pb.Probe_Failed,
+			})
+			continue
+		}
+
+		// pingStats store ping statistics from 0 to 10000 ms (10 seconds)
+		// up to 3 digits precision
+		pingStats := hdrhistogram.New(0, 10000, 3)
```
klizhentas: Why is this histogram created in a loop? I was expecting the check holding the histogram and updating it periodically.

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+	// ping each other node and fail in case the results are over a specified
+	// threshold
+	err = nil
+	for _, node := range nodes {
+		// skip pinging self
+		if node.Addr.String() == selfNode.Addr.String() {
+			continue
+		}
+
+		coord2, err := client.GetCoordinate(node.Name)
```
klizhentas: coord2 is not descriptive name

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+	// finding what is the current node
+	var selfNode serf.Member
+	for _, node := range nodes {
+		// cut the port portion of the address after the ":" away
+		if node.Name == c.serfRPCName {
+			selfNode = node
+		}
+	}
+	if selfNode.Name == "" {
+		log.Errorf("error getting selfNode config: %s", c.serfRPCName)
```
klizhentas: to avoid log.Errorf everywhere, just move all the code to the lower check() error function that can return error, then your check should look like this:

```go
Check() {
   err := f.check()
   if err != nil {
      log.Error(err)
   }
}
```

and you won't have to add log.Error everywhere, just return trace.Wrap

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+		log.Errorf("failed fetching Serf Members - %v", err.Error())
+		r.Add(&pb.Probe{
+			Checker: c.Name(),
+			Status:  pb.Probe_Failed,
+		})
+		return
+	}
+
+	// finding what is the current node
+	var selfNode serf.Member
```
klizhentas: selfNode is verbose, `self` is enough

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+		r.Add(&pb.Probe{
+			Checker: c.Name(),
+			Status:  pb.Probe_Failed,
+		})
+		return
+	}
+	// ping each other node and fail in case the results are over a specified
+	// threshold
+	err = nil
+	for _, node := range nodes {
```
klizhentas: This logic does not conform to the specs, it leads to all nodes pinging each other

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+		log.Errorf("error getting selfNode coordinates: %s -> %#v", selfNode.Name, err)
+		r.Add(&pb.Probe{
+			Checker: c.Name(),
+			Status:  pb.Probe_Failed,
+		})
+		return
+	}
+	// ping each other node and fail in case the results are over a specified
+	// threshold
+	err = nil
```
klizhentas: see my comment about function structure above, this err is not necessary.

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+		// up to 3 digits precision
+		pingStats := hdrhistogram.New(0, 10000, 3)
+		for i := 0; i < slidingWindowSize; i++ {
+			rttNanoSec := selfCoord.DistanceTo(coord2).Nanoseconds()
+			pingStats.RecordValue(rttNanoSec)
+		}
+
+		log.Debugf("%s <-ping-> %s = %d", selfNode.Name, node.Name, pingStats.ValueAtQuantile(pingRttQuantile))
+
+		if pingStats.ValueAtQuantile(pingRttQuantile) >= int64(pingRttThreshold*1e6) { // converting from ms to nanoseconds for comparison
```
klizhentas: magic number, please use constant.

monitoring/ping.go
eldios: changed to `otherNode` which is a little more explicative.

`coord2` is the same naming convention used by the Serf RTT package, I just reused it.

monitoring/ping.go
eldios: changed.

monitoring/ping.go
eldios: I think it does cause the check is only added to Masters, so it will only check values between Master nodes.
I initially added a check but removed because of a comment above.

cmd/agent/checkers.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package main
 
 import (
+	log "github.com/sirupsen/logrus"
```
r0mant: Nitpick: According to the [Go style guide](https://github.com/golang/go/wiki/CodeReviewComments) we follow this import should belong to the 3rd group.

cmd/agent/checkers.go
```diff
@@ -28,6 +30,10 @@ import (
 type config struct {
 	// role is the current agent's role
 	role agent.Role
+	// serfRPCAddr is the Serf RPC endpoint address
+	serfRPCAddr string
+	// serfRPCName is the Serf RPC endpoint address
```
r0mant: Looks like the comment was copied from the above field.

cmd/agent/main.go
```diff
@@ -160,6 +163,7 @@ func run() error {
 			},
 			nettestContainerImage: *cagentNettestContainerImage,
 		}
+		log.Debugf("Agent started using MonitoringConfig %#v", monitoringConfig)
```
r0mant: I think this is the 3rd time I see this config being logged. Do you really need this? :)

monitoring/checkers.go
```diff
@@ -59,6 +59,12 @@ func NodesStatusHealth(config KubeConfig, nodesReadyThreshold int) health.Checke
 	return NewNodesStatusChecker(config, nodesReadyThreshold)
 }
 
+// PingHealth creates a checker that monitors ping values between Master nodes
+// and other nodes
+func PingHealth(serfRPCAddr string, serfRPCName string) health.Checker {
```
r0mant: Nitpick: You could've used shorter form for function arguments:
```go
func PingHealth(serfRPCAddr, serfRPCName string) health.Checker
```

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/gravitational/trace"
+
+	"github.com/codahale/hdrhistogram"
+	serf "github.com/hashicorp/serf/client"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	pingCheckerID     = "ping-checker"
+	slidingWindowSize = 10   // number of ping results to consider per iteration
```
r0mant: Please follow standard Go style for Godocs:

```go
// slidingWindowSize is the number of ping results to consider per iteration
slidingWindowSize = 10
```

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+	pingRttThreshold  = 15.0 // pingRTT threshold expressed in ms
```
r0mant: `Rtt` -> `RTT` here and below.

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+	pingRttQuantile   = 95.0 // quantile used to check against Rtt results
+)
+
+// NewPingChecker implements and return an health.Checker
```
r0mant: Not very useful description. I'd mention what the ping checker does, briefly.

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+	log.Debugf("Using Serf IP: %v", c.serfRPCAddr)
+	log.Debugf("Using Serf Name: %v", c.serfRPCName)
+	clientConfig := serf.Config{
+		Addr: c.serfRPCAddr,
+	}
+	client, err := serf.ClientFromConfig(&clientConfig)
+	if err != nil {
+		log.Errorf("error while connecting to Serf via IP %v. Error: %v",
+			c.serfRPCAddr, err.Error())
+		r.Add(&pb.Probe{
```
r0mant: I would add the error message to the probe, here and everywhere below where the error happens. There is `NewProbeFromErr` function IIRC.

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+			Status:  pb.Probe_Failed,
+		})
+		return
+	}
+	defer client.Close()
+
+	// retrieve other nodes using Serf members
+	nodes, err := client.Members()
+	if err != nil {
+		log.Errorf("failed fetching Serf Members - %v", err.Error())
```
r0mant: Instead of logging the error every time it happens, I would move the check out to a separate method that returns an error, and the `Check` function would just call it and log the error once (and create a failed probe):

```go
func (c *pingChecker) Check(ctx context.Context, r health.Reporter) {
  err := c.check(ctx, r)
  if err != nil {
    // log the error and create a failed probe
  }
}

func (c *pingChecker) check(ctx context.Context, r health.Reporter) error {
  // does what this function does but returns errors and doesn't log them
}
```

monitoring/ping.go
r0mant: @eldios Yeah, you can store `map[node]histogram` on the pingChecker and feed values to appropriate histogram on every iteration.

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+		pingStats := hdrhistogram.New(0, 10000, 3)
+		for i := 0; i < slidingWindowSize; i++ {
+			rttNanoSec := selfCoord.DistanceTo(coord2).Nanoseconds()
+			pingStats.RecordValue(rttNanoSec)
+		}
+
+		log.Debugf("%s <-ping-> %s = %d", selfNode.Name, node.Name, pingStats.ValueAtQuantile(pingRttQuantile))
+
+		if pingStats.ValueAtQuantile(pingRttQuantile) >= int64(pingRttThreshold*1e6) { // converting from ms to nanoseconds for comparison
+			log.Errorf("slow ping between nodes detected. Value %v over threshold %v",
```
r0mant: I'd maybe log it as a warning, not error.

monitoring/ping.go
```diff
@@ -0,0 +1,179 @@
...
+	if selfNode.Name == "" {
+		log.Errorf("error getting selfNode config: %s", c.serfRPCName)
+		r.Add(&pb.Probe{
+			Checker: c.Name(),
+			Status:  pb.Probe_Failed,
+		})
+		return
+	}
+
+	selfCoord, err := client.GetCoordinate(selfNode.Name)
```
r0mant: To make the code more structured, I would move the logic of calculating RTT b/w two nodes in a separate method. E.g.:

```go
func calculateRTT(serfClient ..., node1, node2 ...) (int64, error) {  // or whatever type it returns
}
```

Then your check will simplify:

```go
for _, node := range nodes {
  rtt, err := calculateRTT(serdClient, selfNode, node)
  // add rtt to histogram, check threshold, etc.
}
```

monitoring/ping.go
```diff
@@ -0,0 +1,189 @@
...
+			continue
+		}
+
+		// pingStats store ping statistics from 0 to 10000 ms (10 seconds)
+		// up to 3 digits precision
+		pingStats := hdrhistogram.New(0, 10000, 3)
+		for i := 0; i < slidingWindowSize; i++ {
+			rttNanoSec := selfCoord.DistanceTo(coord2).Nanoseconds()
+
+			if c.rttStats[node.Name] == nil { // FIXME: NOT WORKING
```
r0mant: You should initialize histograms when the pingChecker is created, then you won't need this.

cmd/agent/checkers.go
eldios: fixed.

monitoring/checkers.go
eldios: nice!

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+			continue
+		}
+
+		// pingStats store ping statistics from 0 to 10000 ms (10 seconds)
+		// up to 3 digits precision
+		pingStats := hdrhistogram.New(0, 10000, 3)
+		for i := 0; i < slidingWindowSize; i++ {
+			rttNanoSec := selfCoord.DistanceTo(coord2).Nanoseconds()
+
+			_, exists := c.rttStats[node.Name]
```
klizhentas: what happens when the node gets removed? are we leaking histograms?

eldios: Good point. :thinking: 
Those values are stored in memory, so they'll go away with the node itself.

It's true though that a trace of all those nodes will remain in all nodes' memory until they will also restart and/or leave.

monitoring/ping.go
eldios: Fixed.

monitoring/ping.go
a-palchikov: I think the comment above has been ignored.

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+
+// Check verifies that all nodes' ping with Master Nodes is lower than the
+// desired threshold
+// Implements health.Checker
+func (c *pingChecker) Check(ctx context.Context, r health.Reporter) {
+	// FIXME: #1 RttThreshold will become configurable in future
+	// FIXME: #2 Send RttThreshold value to metrics
+
+	// fetch serf config and intantiate client
+	log.Debugf("Using Serf IP: %v", c.serfRPCAddr)
```
a-palchikov: That's not going to be useful if output on each test step.

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+			Checker: c.Name(),
+			Status:  pb.Probe_Failed,
+		})
+		return
+	}
+
+	// finding what is the current node
+	var self serf.Member
+	for _, node := range nodes {
+		// cut the port portion of the address after the ":" away
```
a-palchikov: Is this up to date since the comment does not reflect what the code does?

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+		coord2, err := client.GetCoordinate(node.Name)
+		if err != nil {
+			log.Errorf("error getting coordinates: %s -> %#v", node.Name, err)
+			r.Add(&pb.Probe{
+				Checker: c.Name(),
+				Status:  pb.Probe_Failed,
+			})
+			continue
+		}
+		if coord2 == nil {
```
a-palchikov: Can you elaborate what it means when a coordinate is not available?

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+			r.Add(&pb.Probe{
+				Checker: c.Name(),
+				Status:  pb.Probe_Failed,
+			})
+			continue
+		}
+
+		// pingStats store ping statistics from 0 to 10000 ms (10 seconds)
+		// up to 3 digits precision
+		pingStats := hdrhistogram.New(0, 10000, 3)
```
a-palchikov: `10000` is not readable at all.
Create a const for a second and use `10 * second`.

monitoring/ping.go
a-palchikov: I'm also curious how the coordinates are determined / refreshed. What should happen if one (or all) are not available (i.e. `GetCoordinate()` returns `nil` for coordinate and a `nil` error)?

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+				Checker: c.Name(),
+				Status:  pb.Probe_Failed,
+			})
+		} else {
+			log.Debugf("ping value %v below threshold %v ms", pingStats.ValueAtQuantile(pingRttQuantile), pingRttThreshold)
+		}
+	}
+
+	// set Probe to be running
+	if err != nil {
```
a-palchikov: Wrong condition.

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+
+		log.Debugf("%s <-ping-> %s = %d", self.Name, node.Name, pingStats.ValueAtQuantile(pingRttQuantile))
+
+		if pingStats.ValueAtQuantile(pingRttQuantile) >= int64(pingRttThreshold*msToNanoSec) {
+			log.Errorf("slow ping between nodes detected. Value %v over threshold %v",
+				pingRttQuantile, pingRttThreshold)
+			r.Add(&pb.Probe{
+				Checker: c.Name(),
+				Status:  pb.Probe_Failed,
+			})
```
a-palchikov: Also, explicit `return` is missing here.

cmd/agent/checkers.go
eldios: fixed.

cmd/agent/main.go
eldios: No I was trying to understand where was the configuration lost and added similar debugging statements in multiple places.

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+	"github.com/gravitational/trace"
+
+	"github.com/codahale/hdrhistogram"
+	serf "github.com/hashicorp/serf/client"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	pingCheckerID               = "ping-checker" // pingCheckerID specifies the check name
```
klizhentas: usually, we place comments on the line above the variable, because Go can process them in godoc.

monitoring/ping.go
```diff
@@ -0,0 +1,191 @@
...
+	slidingWindowSize           = 10             // slidingWindowSize specifies the sliding window use by checks
+	pingRttMin                  = 0              // pingRttMin set the minim value that can be recorded
```
klizhentas: Rtt is not clear, `pingRoundtripMinimum` is more clear.

monitoring/ping.go
r0mant: Re: removed nodes - consider using github.com/gravitational/ttlmap with TTL of, say, 1h. This way the histogram of a removed node will be gone in 1h. The only inconvenience is that with TTL map you'd have to re-set the node's histogram to the map on each check to refresh its TTL.

monitoring/ping.go
eldios: Fixed.

cmd/agent/checkers.go
```diff
@@ -28,6 +30,10 @@ import (
 type config struct {
 	// role is the current agent's role
 	role agent.Role
+	// serfRPCAddr is the Serf RPC endpoint address
+	serfRPCAddr string
+	// serfRPCName is the Node name used in the Serf cluster as the member name
```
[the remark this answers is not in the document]
eldios: fixed.

cmd/agent/main.go
```diff
@@ -146,8 +146,11 @@ func run() error {
 			CertFile:    *cagentCertFile,
 			KeyFile:     *cagentKeyFile,
 		}
+		log.Debugf("Agent started using agentConfig %#v", agentConfig)
```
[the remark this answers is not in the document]
eldios: fixed.

monitoring/ping.go
eldios: fixed.

klizhentas:
To iterate on what Roman said about gravitational/ttlmap:

* Create a ttlmap to hold histograms with ttl of 1 hour.
* If you receive ping, update TTL on the existing histogram to prevent expriy and data erasure
* Histograms from deleted nodes will be wiped out in an hour.

Otherwise, you risk to have accumulated histograms over period of time in case of auto scaling groups.

Access to the map should be protected by sync.Mutex, not sync.RWMutex because ttlmap.Get modifies the map (expires it).

monitoring/ping.go
```diff
@@ -160,7 +160,7 @@ func (c *pingChecker) checkNodesRTT(nodes []serf.Member, client *serf.RPCClient)
 		}
 
 		rttStatsHDRInterface, _ := c.rttStats.Get(node.Name)
-		rttStatsHDR := rttStatsHDRInterface.(hdrhistogram.Histogram)
+		rttStatsHDR := rttStatsHDRInterface.(*hdrhistogram.Histogram)
```
klizhentas: this should not panic, so should be `rttStatsHDR, ok :=` and next line `if !ok {return err}`.

Also `rttStatsHDR` was hard for me to read, can we rename it to something like `roundtripLatency`. You don't need to communicate the type of the variable, because Go does this automatically - it's known that it's a Histogram.

monitoring/ping.go
```diff
@@ -183,29 +188,38 @@ func (c *pingChecker) checkNodesRTT(nodes []serf.Member, client *serf.RPCClient)
...
-		nodeTTLMapInterface, _ = c.rttStats.Get(node.Name)
+		nodeTTLMapInterface, _ = c.roundtripLatency.Get(node.Name)
 	}
 
-	nodeTTLMap := nodeTTLMapInterface.(*hdrhistogram.Histogram)
+	nodeTTLMap, ok := nodeTTLMapInterface.(*hdrhistogram.Histogram)
+	if !ok {
+		errMsg := fmt.Sprintf("couldn't parse roundtripLatency as HDRHistogram on %s",
+			c.serfMemberName)
+		return errors.New(errMsg)
```
klizhentas: please use `trace.BadParameter("... %v", ...)` here instead of errors package here and everywhere else

monitoring/ping.go
```diff
@@ -183,29 +188,38 @@ func (c *pingChecker) checkNodesRTT(nodes []serf.Member, client *serf.RPCClient)
...
 			hdrhistogram.New(pingRoundtripMinimum.Nanoseconds(),
 				pingRoundtripMaximum.Nanoseconds(),
 				pingRoundtripSignificativeFigures),
 			statsTTLPeriod)
-		nodeTTLMapInterface, _ = c.rttStats.Get(node.Name)
+		nodeTTLMapInterface, _ = c.roundtripLatency.Get(node.Name)
 	}
 
-	nodeTTLMap := nodeTTLMapInterface.(*hdrhistogram.Histogram)
+	nodeTTLMap, ok := nodeTTLMapInterface.(*hdrhistogram.Histogram)
```
klizhentas: I have a same comment here, in Go there is no need to communicate the type of the variable via variable name, because Go is a statically typed language. Instead it's better to communicate the intent of the variable:

So `nodeTTLMap` should be `nodeLatencies` and `nodeTTLMapInterface` should a temporary variable `n` because it does not carry any meaning and is used in one line only

monitoring/ping.go
```diff
@@ -0,0 +1,263 @@
...
+		// pop element at index 0 (oldest)
+		nodeLatencies.Reset()
+
+		lowerLimit := len(tmpSnapshot.Counts) - slidingWindowSize
+		if lowerLimit < 0 {
+			lowerLimit = 0
+		}
+		// using len(tmpSnapshot.Counts)-1 as the upper limit as another value
+		// will be added a few lines below
+		for i := lowerLimit; i < len(tmpSnapshot.Counts)-1; i++ {
```
r0mant: Hmm, I don't think this has an effect that you indented it to have. If you look at how the hdrhistogram is implemented, you'll see that it stores values in a "sparse" list ("counts" field you use here) - when you record a value in it, it calculates the value's "index" and then increments the value in the list at that index by 1. So iterating over "counts" here does not actually mean iterating over values oldest-to-newest.

If you want to keep a sliding window of values, I think that instead of storing histogram in the map, you will need to store a list of values as they arrive and then construct histogram on demand in the check. I.e. (semi-pseudo-Go):

```go
type pingChecker struct {
    windows TTLMap // TTL map where key is node name and value is `[]int64`
}

func (c *pingChecker) storePingValue(node, value) error {
  // first record value for the node in the list, throwing out the older element if needed
  latencies, ok := c.windows.Get(node) // latencies is a slice of latencies
  // see if we need to truncate the list
  if len(latencies) >= maxWindowSize {
    latencies = latencies[1:]
  }
  // append new value
  latencies = append(latencies, value)
  // build hdr histogram
  histogram := hdrhistogram.New(...)
  for _, latency := range latencies {
    histogram.RecordValue(latency)
  }
  // check percentiles, etc.
}
```
Let me know if that makes sense.

monitoring/ping.go
```diff
@@ -0,0 +1,263 @@
...
+		s, _ = c.roundtripLatency.Get(node.Name)
+	}
+
+	nodeLatencies, ok := s.(*hdrhistogram.Histogram)
+	if !ok {
+		return trace.BadParameter("couldn't parse roundtripLatency as HDRHistogram on %s", c.serfMemberName)
+	}
+
+	tmpSnapshot := nodeLatencies.Export()
+	if len(tmpSnapshot.Counts) >= slidingWindowSize {
```
a-palchikov: I would simplify like this:
```go
snapshot := nodeLatencies.Export()
if len(snapshot.Counts) > slidingWindowSize {
	nodeLatencies.Reset()
	// record the values outside the window
	for _, latency := range snapshot.Counts[slidingWindowSize:] {
		nodeLatencies.RecordValue(latency)
	}
}
```

monitoring/ping.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	log.Debugf("%d recorded ping RoundTrip values for node %s",
+		nodeLatencies.TotalCount(), node.Name)
+
+	return nil
+}
+
+// calculateRTT calculates the RoundTrip time between two Serf Cluster members
+func calculateRTT(serfClient *serf.RPCClient, self serf.Member, node serf.Member) (int64, error) {
```
a-palchikov: nit: you can combine arguments of the same type:
```go
func calculateRTT(serfClient *serf.RPCClient, self, node serf.Member) (rttNanos int64, err error) {
```
Also note that it's recommended to name the returns for readability.

monitoring/ping.go
```diff
@@ -0,0 +1,263 @@
...
+	if selfCoord == nil {
+		return 0, trace.NotFound("self node %s coordinates not found", self.Name)
+	}
+
+	otherNodeCoord, err := serfClient.GetCoordinate(node.Name)
+	if err != nil {
+		return 0, trace.NotFound("error getting coordinates: %s -> %v", node.Name, err)
+	}
+	if otherNodeCoord == nil {
+		return 0, trace.NotFound("could not find a coordinate for node %s -> %v", node.Name, err)
```
a-palchikov: `err` will always be `nil` here

monitoring/ping.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	return selfCoord.DistanceTo(otherNodeCoord).Nanoseconds(), nil
+}
+
+// setProbeStatus set the Probe according to status or raise an error if one is passed via arguments
```
a-palchikov: set**s** ... raise**s**
You don't need this method - it does not provide anything beyond what's already available with `NewProbeFromErr`.

monitoring/checkers.go
```diff
@@ -59,6 +59,12 @@ func NodesStatusHealth(config KubeConfig, nodesReadyThreshold int) health.Checke
 	return NewNodesStatusChecker(config, nodesReadyThreshold)
 }
 
+// PingHealth creates a checker that monitors ping values between Master nodes
+// and other nodes
+func PingHealth(serfRPCAddr, serfMemberName string) health.Checker {
+	return NewPingChecker(serfRPCAddr, serfMemberName)
```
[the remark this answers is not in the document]
eldios: tbh I was just following the same pattern used by all the other checks.

If it's not the desired approach I can definitely go straight to NewPingChecker()

monitoring/ping.go
```diff
@@ -0,0 +1,263 @@
...
+		return trace.BadParameter("couldn't parse roundtripLatency as HDRHistogram on %s", c.serfMemberName)
+	}
+
+	tmpSnapshot := nodeLatencies.Export()
+	if len(tmpSnapshot.Counts) >= slidingWindowSize {
+		// pop element at index 0 (oldest)
+		nodeLatencies.Reset()
+
+		lowerLimit := len(tmpSnapshot.Counts) - slidingWindowSize
+		if lowerLimit < 0 {
```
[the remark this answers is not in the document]
eldios: seems like an old left-over :thinking: 
Removed.

monitoring/ping.go
eldios: in a comment above both you and Roman told me to not use log.Error() everywhere and instead wrap it inside a dedicated function... that's why this function is here now.

monitoring/ping.go
eldios: makes sense, I'll reimplement it following your suggestion.

monitoring/ping.go
```diff
@@ -0,0 +1,263 @@
...
+	pingRoundtripSignificativeFigures = 3
+	// pingRoundtripThreshold sets the RTT threshold
+	pingRoundtripThreshold = 15 * time.Millisecond
+	// pingRoundtripQuantile sets the quantile used while checking Histograms against Rtt results
+	pingRoundtripQuantile = 95.0
+)
+
+// NewPingChecker implements and return an health.Checker
+func NewPingChecker(serfRPCAddr string, serfMemberName string) health.Checker {
+	roundtripLatencyTTLMap, err := ttlmap.New(int(statsTTLPeriod.Seconds()))
```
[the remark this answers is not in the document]
eldios: I got it wrong and thought it was the maximum TTL possible.

I'm not sure what to set here though cause this would be the number of "other" nodes in the cluster but this number can vary over time and there's no upper limit that I know of in Gravity.

Maybe I could connect to Serf and fetch the current number of nodes and use that... but then again it may change over time :man_shrugging:

I guess I can use the number from Serf initially and then update it later before Set() calls?

monitoring/ping.go
a-palchikov: yes, but in this case you could simply rewrite top-level `Check` like this:
```go
func (c *pingChecker) Check(ctx context.Context, r health.Reporter) {
	// FIXME: #1 RttThreshold will become configurable in future
	// FIXME: #2 Send RttThreshold value to metrics

	err := c.check(ctx, r)
	if err != nil {
		log.Warn(err.Error())
		r.Add(NewProbeFromErr(c.Name(), "", err))
		return
	}
	r.Add(NewSuccessProbe(c.Name()))
}
```

monitoring/ping.go
```diff
@@ -0,0 +1,289 @@
...
+)
+
+// NewPingChecker returns a checker that verifies accessibility of nodes in the cluster by exchanging ping requests
+func NewPingChecker(serfRPCAddr string, serfMemberName string) (c health.Checker, err error) {
+	latencyTTLMap, err := ttlmap.New(int(statsTTLPeriod.Seconds()))
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	log.Debugf("[ping] using Serf IP: %v", serfRPCAddr)
```
klizhentas: Instead of `[ping]` use log.WithFields that will have the same outcome

monitoring/ping.go
```diff
@@ -230,7 +230,7 @@ func (c *pingChecker) saveLatencyStats(pingLatency int64, node serf.Member) erro
 
 	s, exists := c.latencyStats.Get(node.Name)
 	if !exists {
-		sMap = make([]int64, slidingWindowSize)
+		sMap = make([]int64, 0, slidingWindowSize)
```
klizhentas: why is this slice called `sMap`?

eldios: initial it was actually a `map` then since a Slice is just a map between int and something else the name remained. Anyway I fixed it now.

r0mant: @eldios That's interesting way to think about a slice, although I'm afraid incorrect :) Slice is backed by array and array and map are two completely different data structures with different properties.

monitoring/ping.go
```diff
@@ -219,14 +221,18 @@ func (c *pingChecker) buildLatencyHistogram(nodeName string) (latencyHDR *hdrhis
...
+		}
 	}
 
 	return latencyHDR, nil
 }
 
 // saveLatencyStats is used to store ping values in HDR Histograms in memory
 func (c *pingChecker) saveLatencyStats(pingLatency int64, node serf.Member) error {
+	c.mux.Lock()
 	var latencySlice []int64
```
klizhentas: just add `defer c.mux.Unlock()` on this line and remove all other unlock calls

monitoring/ping.go
```diff
@@ -0,0 +1,308 @@
...
+	if err != nil {
+		return 0, trace.Wrap(err)
+	}
+	if selfCoord == nil {
+		return 0, trace.NotFound("self node %s coordinates not found", self.Name)
+	}
+
+	otherNodeCoord, err := serfClient.GetCoordinate(node.Name)
+	if err != nil {
+		return 0, trace.NotFound("error getting coordinates: %s -> %v", node.Name, err)
```
a-palchikov: Should use `trace.Wrap` - as in the self case above.

monitoring/ping.go
```diff
@@ -0,0 +1,308 @@
...
+	} else {
+		var ok bool
+		latencySlice, ok = s.([]int64)
+		if !ok {
+			c.mux.Unlock()
+			return trace.BadParameter("couldn't parse node latency as []int64 on %s", c.serfMemberName)
+		}
+
+		// upperLimit needs to be the highest between `len(latencySlice)` and `slidingWindowSize`
+		upperLimit := max(len(latencySlice), slidingWindowSize)
```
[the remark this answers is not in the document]
eldios: There's no harm in checking => as opposed to ==, so I guess it's better being on the safe side.

Besides bugs and hiccups slidingWindowSize should never end up being > that latencies (latencySlice) though.

monitoring/ping.go
```diff
@@ -0,0 +1,308 @@
...
+	log.Debugf("using Serf Name: %v", serfMemberName)
+	// fetch serf config and instantiate client
+	clientConfig := serf.Config{
+		Addr: serfRPCAddr,
+	}
+	client, err := serf.ClientFromConfig(&clientConfig)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	// FIXME: defer client.Close() # cannot be used here or causes error: ping-checker (set tcp 127.0.0.1:52026: use of closed network connection)
```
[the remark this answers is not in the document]
eldios: thought I had to avoid leaking connections handlers and leaving them open.
But you're right and I got it wrong, so removing this.

monitoring/ping.go
```diff
@@ -0,0 +1,308 @@
...
+	latencyTTLMap, err := ttlmap.New(int(statsTTLPeriod.Seconds())) //FIXME: why is number of seconds used as capacity?
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	log.WithFields(log.Fields{trace.Component: "ping"})
+	log.Debugf("using Serf IP: %v", serfRPCAddr)
+	log.Debugf("using Serf Name: %v", serfMemberName)
+	// fetch serf config and instantiate client
+	clientConfig := serf.Config{
```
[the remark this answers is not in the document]
eldios: I don't think we should implement ALL features now, otherwise this task will never end and will be absorbed in a never ending whirlwind of scope-creep.

I can definitely add it as a TODO comment, along with the other desired features but I don't think it should be implemented now if we don't have a clear, pragmatic, real application that we use right now.

monitoring/ping.go
```diff
@@ -0,0 +1,314 @@
...
+		if latencies, ok = value.([]int64); !ok {
+			return nil, trace.BadParameter("couldn't parse node latency as []int64 on %s", c.serfMemberName)
+		}
+	}
+
+	if len(latencies) >= slidingWindowSize {
+		// pop oldest value to make room for the new one
+		copy(latencies, latencies[1:])
+		// keep the slice within the sliding window size
+		latencies = latencies[:slidingWindowSize-1]
```
r0mant: Why `-1`? You're losing the last element. Also, why do you need "copy" first? I would simplify all this to just:

```go
if len(latencies) >= slidingWindowSize {
  latencies = latencies[1:slidingWindowSize] // or event just latencies = latencies[1:] since it's never gonna exceed it
}
```

monitoring/ping.go
```diff
@@ -0,0 +1,314 @@
...
+	"github.com/gravitational/ttlmap"
+	serf "github.com/hashicorp/serf/client"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	// pingCheckerID specifies the check name
+	pingCheckerID = "ping-checker"
+	// slidingWindowSize specifies the number of retained check results
+	slidingWindowSize = 10
```
r0mant: I think this window size is too small to indicate anything and might produce a lot of false positives, let's increase to a 100 at least.

monitoring/ping.go
```diff
@@ -0,0 +1,314 @@
...
+	return
+}
+
+// check runs the actual system status verification code and returns an error
+// in case issues arise in the process
+func (c *pingChecker) check(ctx context.Context, r health.Reporter) (err error) {
+
+	client := &c.serfClient
+	// check if client connection closed and reopen it
+	if client.IsClosed() {
```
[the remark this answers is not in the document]
eldios: This comes from a comment @a-palchikov added about a case in which the Serf connection is unstable (or the cluster is restarted) and the connection gets closed.
Thus the check plus the client reopening.

monitoring/ping.go
eldios: 10 was the proposed number initially. It seems that the check runs every 30 seconds so 100 would take a lot of time to change the 95 percentile in case things go bad, I think.
I guess it's a tradeoff between being reactive and dealing with false positives by tweaking the threshold and the slidingWindow.
I'll put 100 for now but it seems a little too  wide as a window IMO.

monitoring/ping.go
eldios: slidingWindowSize needs to have -1 because another element will be appended a few lines below as the latest/newest element.

Removed copy, it was there from a proposal that @a-palchikov did but I guess it's redundant now.

monitoring/ping.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+
+		log.Debugf("%s <-ping-> %s = %dns [latest]", c.self.Name, node.Name, rttNanoSec)
+		log.Debugf("%s <-ping-> %s = %dns [%.2f percentile]",
+			c.self.Name, node.Name,
+			latencyHistogram.ValueAtQuantile(latencyQuantile),
+			latencyQuantile)
+
+		latencyPercentile := latencyHistogram.ValueAtQuantile(latencyQuantile)
+		if latencyPercentile >= latencyThreshold.Nanoseconds() {
```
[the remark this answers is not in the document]
eldios: I guess that's bound to interpretation. Since it's a percentile I'd be conservative and start alerting as soon as it reaches the threshold... not when it's strictly over the threshold.

But then again it's really bound to interpretation so it can be both ways, I guess? :thinking: :man_shrugging:

monitoring/dns.go
```diff
@@ -29,19 +29,54 @@ import (
 type DNSChecker struct {
 	// QuestionA is a slice of questions to ask for A (Host) records
 	QuestionA []string
+	// QuestionNS is a slice of questions to ask for NS (Nameserver) records
```
a-palchikov: I have an off-topic question about how these changes ended up here?

monitoring/ping.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func newSerfClient(serfRPCAddr string) (client *serf.RPCClient, err error) {
+	// fetch serf config and instantiate client
+	clientConfig := serf.Config{
+		Addr: serfRPCAddr,
+	}
+	client, err = serf.ClientFromConfig(&clientConfig)
+	if err != nil {
+		return client, trace.Wrap(err)
```
a-palchikov: nit: `client` is `nil` here - just return an explicit `nil`

monitoring/ping.go
```diff
@@ -0,0 +1,283 @@
...
+		var ok bool
+		if latencies, ok = value.([]int64); !ok {
+			return nil, trace.BadParameter("couldn't parse node latency as []int64 on %s", c.serfMemberName)
+		}
+	}
+
+	if len(latencies) >= latencyStatsSlidingWindowSize {
+		// keep the slice within the sliding window size
+		// slidingWindowSize is -1 because another element will be added a few lines below
+		latencies = latencies[1 : latencyStatsSlidingWindowSize-1]
```
a-palchikov: this will override the last new value. The range is not right-inclusive - the last remaining element will be `latencies[latencyStatsSlidingWindowSize-2]`. You need to:
```go
latencies = latencies[1 : latencyStatsSlidingWindowSize]
```
nit: also, in the suggested version the slice reused the same backing array which stayed `latencyStatsSlidingWindowSize` in size - this implementation will be constantly reallocating (`latencyStatsSlidingWindowSize * 2` in worst-case).

monitoring/dns.go
r0mant: @eldios Yeah, actually I'm kinda curious about it too. Seems unrelated to the ping check?

monitoring/ping.go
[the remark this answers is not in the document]
r0mant: @eldios @a-palchikov Isn't serf client self-reconnecting though? So if connection was closed, it will detect it and reconnect on the `Members` call below? https://github.com/gravitational/satellite/blob/master/agent/serf.go#L56

agent/serf.go
```diff
@@ -38,9 +39,13 @@ type serfClient interface {
 	Join(peers []string, replay bool) (int, error)
 	// UpdateTags will modify the tags on a running serf agent
 	UpdateTags(tags map[string]string, delTags []string) error
+	// TODO
+	GetCoordinate(node string) (*coordinate.Coordinate, error)
 }
 
-func newSerfClient(clientConfig serf.Config) (*retryingClient, error) {
+type NewSerfClientFunc func(serf.Config) (SerfClient, error)
```
r0mant: Can you please add godocs on all exported types/functions in this module?

agent/server_test.go
```diff
@@ -29,6 +29,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/serf/coordinate"
```
r0mant: Nit: This import should go to the 3rd group.

monitoring/ping.go
```diff
@@ -0,0 +1,295 @@
...
+	self           serf.Member
+	serfClient     agent.SerfClient
+	serfRPCAddr    string
+	serfMemberName string
+	latencyStats   ttlmap.TTLMap
+	mux            sync.Mutex
+	logger         log.Entry
+}
+
+type PingCheckerConfig struct {
```
r0mant: Godocs on the struct and its fields please.

monitoring/ping.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// check runs the actual system status verification code and returns an error
+// in case issues arise in the process
+func (c *pingChecker) check(ctx context.Context, r health.Reporter) (err error) {
+
+	client := c.serfClient
+
+	// retrieve other nodes using Serf members
+	nodes, err := client.Members()
```
r0mant: I'm curious - how does this handle removed nodes? I.e. when you shrink a cluster, serf member stays in the "left" state so will it be returned here? And we'll keep trying to measure latency to it? I think we should test this scenario and maybe filter out all non-active members here.

agent/serf.go
```diff
@@ -38,9 +39,15 @@ type serfClient interface {
 	Join(peers []string, replay bool) (int, error)
 	// UpdateTags will modify the tags on a running serf agent
 	UpdateTags(tags map[string]string, delTags []string) error
+	// GetCoordinate get&returns the Serf Coordinate for a specific Node
```
a-palchikov: ~~get&~~ returns a serf cluster coordinate for the specified node

agent/serf.go
```diff
@@ -38,9 +39,15 @@ type serfClient interface {
...
+	GetCoordinate(node string) (*coordinate.Coordinate, error)
 }
 
-func newSerfClient(clientConfig serf.Config) (*retryingClient, error) {
+type NewSerfClientFunc func(serf.Config) (SerfClient, error)
+
+// NewSerfClient is an helper function used to provide a custom Serf Client
```
a-palchikov:
```go
// NewSerfClient returns a new serf client for the specified configuration.
// The client will attempt to reconnect if it detects if the connection to the
// serf agent has been lost
```

agent/serf.go
```diff
@@ -103,6 +110,16 @@ func (r *retryingClient) Close() error {
 	return r.client.Close()
 }
 
+// GetCoordinate get&returns the Serf Coordinate for a specific node
```
a-palchikov: See comment on the API method above.

agent/serf.go
```diff
@@ -131,3 +148,49 @@ type retryingClient struct {
 	client *serf.RPCClient
 	config serf.Config
 }
+
+// MockSerfClient is a mock/fake Serf Client used in testing
```
a-palchikov: This belongs in a test file - i.e. serf_test.go

agent/serf.go
```diff
@@ -131,3 +148,49 @@ type retryingClient struct {
...
+type MockSerfClient struct {
+	members []serf.Member
+	coords  map[string]*coordinate.Coordinate
+}
+
+// NewMockSerfClient is an helper function used to create a mock/fake Serf Client
```
a-palchikov: is **a** helper that creates a new instance of the mock client.

agent/serf.go
```diff
@@ -131,3 +148,49 @@ type retryingClient struct {
...
+// used in testing
+func NewMockSerfClient(members []serf.Member, coords map[string]*coordinate.Coordinate) (client *MockSerfClient, err error) {
+	return &MockSerfClient{
+		members: members,
+		coords:  coords,
+	}, nil
+}
+
+// Members is a function that return the (fake) Serf member nodes
```
a-palchikov: that return**s** ... You can remove `fake` (as it will be obvious from the test context) when you move this implementation to serf_test.go

agent/server_test.go
```diff
@@ -657,6 +657,10 @@ func (r *testSerfClient) Join(peers []string, replay bool) (int, error) {
 	return 0, nil
 }
 
+func (r *testSerfClient) GetCoordinate(node string) (coord *coordinate.Coordinate, err error) {
```
a-palchikov: I don't quite understand the reason for both testSerfClient and MockFakeSerfClient - can there be just one for testing?

monitoring/ping_test.go
```diff
@@ -0,0 +1,126 @@
...
+			Coords: map[string]*coordinate.Coordinate{
+				"member-1": &coordinate.Coordinate{
+					Height: 1,
+				},
+				"member-2": &coordinate.Coordinate{
+					Height: 1,
+				},
+			},
+			Status:      agentpb.NodeStatus_Running,
+			Description: "Testing nodes with failing status",
```
a-palchikov: Why is the expected status `NodeStatus_Running`?

monitoring/ping_test.go
```diff
@@ -0,0 +1,126 @@
...
+		},
+	}
+
+	for _, testCase := range testCases {
+		pingChecker, err := NewPingChecker(PingCheckerConfig{
+			SerfRPCAddr:    "127.0.0.1",
+			SerfMemberName: "member-1",
+			NewSerfClient: func(serf.Config) (agent.SerfClient, error) {
```
a-palchikov: Instead of config.SerfRPCAddr/NewSerfClient consider passing an initialized serf client to checker.

monitoring/ping.go
```diff
@@ -0,0 +1,306 @@
...
+	serfRPCAddr    string
+	serfMemberName string
+	latencyStats   ttlmap.TTLMap
+	mux            sync.Mutex
+	logger         log.Entry
+}
+
+// PingCheckerConfig is used to store all the configuration related to the current check
+type PingCheckerConfig struct {
+	// SerfRPCAddr is the address used by the Serf RPC client to communicate
```
a-palchikov: nit: the comment reads incomplete - to communicate with what?

monitoring/ping.go
```diff
@@ -0,0 +1,306 @@
...
+	SerfRPCAddr string
+	// SerfMemberName is the name associated to this node in Serf
```
a-palchikov: associated -> assigned

monitoring/ping.go
```diff
@@ -0,0 +1,306 @@
...
+	SerfMemberName string
+	// NewSerfClient is an optional Serf Client function that can be used instead
+	// of the default one. If not specified it will fallback to the default one
+	NewSerfClient agent.NewSerfClientFunc
```
a-palchikov: Have you considered passing the client as part of configuration?

monitoring/ping.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// CheckAndSetDefaults is an helper function which just check that the provided
```
a-palchikov: // CheckAndSetDefaults validates that this configuration is correct and sets value defaults where necessary

monitoring/ping_test.go
eldios: Unless it's a huge mistake, enormous waste of resource or something that breaks other features let's postpone this improvement to a future moment when it will be needed and be done with this task, IMO.

monitoring/ping.go
eldios: ...see comment above.

monitoring/ping.go
```diff
@@ -0,0 +1,306 @@
...
+	return nil
+}
+
+// checkNodesRTT implements the bulk of the logic by checking the ping time
+// between this node (self) and the other Serf Cluster member nodes
+func (c *pingChecker) checkNodesRTT(nodes []serf.Member, client agent.SerfClient) error {
+	// ping each other node and fail in case the results are over a specified
+	// threshold
+	for _, node := range nodes {
+		// skip pinging self
```
a-palchikov: One more thing that occurred to me is that there may be failed/left members in the cluster which should deserve special treatment. At least, left members should not be considered. Also, I don't believe it makes much sense to include failed members in the round trip time computation.

monitoring/ping.go
```diff
@@ -0,0 +1,307 @@
...
+			break // self node found, breaking out of the for loop
+		}
+	}
+	if self.Name == "" {
+		return nil, trace.NotFound("failed to find Serf member with name %s", conf.SerfMemberName)
+	}
+
+	return &pingChecker{
+		self:           self,
+		serfClient:     client,
```
a-palchikov: This seems to be unused.

monitoring/ping.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	// finding what is the current node
+	var self serf.Member
+	for _, node := range nodes {
+		logger.Debugf("node %s status %s", node.Name, node.Status)
+		if node.Status != pb.MemberStatus_Alive.String() {
+			continue
+		}
```
a-palchikov: I don't think this is possible at this early stage - i.e. when the checkers are first added to the agent. I think at this point the agent is not yet part of the serf cluster.

monitoring/ping_test.go
a-palchikov: My point is that this indeed not needed now - i.e. complicating the design with NewSerfClient when one could be just created by the caller and passed on, no?

monitoring/ping.go
eldios: I added this additional check on line 130 and forgot to add it here too :man_facepalming: 

Fixed now.

monitoring/ping_test.go
eldios: ah okay. Yeah I get you're saying.

We went that way so that it would be easier to write tests... and it's done now so I won't bother changing it again. But I agree that it would have been probably easier to go the other way eventually.

eldios:
> @eldios You can merge after you've made sure that:
> 
> * checker can be added to the list of other checkers in agent w/o errors

```
Apr 08 17:43:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      skipping node 172_28_128_133.gra because status is 'failed' monitoring/ping.go:193
Apr 08 17:43:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      172_28_128_131.gra <-ping-> 172_28_128_132.gra = ping okay. Value 447743ns within threshold 15ms (15000000ns) monitoring/ping.go:229
Apr 08 17:43:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      172_28_128_131.gra <-ping-> 172_28_128_132.gra = 447743ns [95.00 percentile] monitoring/ping.go:189
Apr 08 17:43:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      172_28_128_131.gra <-ping-> 172_28_128_132.gra = 447689ns [latest] monitoring/ping.go:189
Apr 08 17:43:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      20 recorded ping values for node 172_28_128_132.gra => [387917 143582 71402 97002 302909 277768 221460 412257 368043 113817 162726 22363 45880 254161 323595 108157 990688 2297
Apr 08 17:43:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      self {[0.0013256854355661921 -0.00014074398739693657 -0.0002627643992015428 -0.0007953835202403196 -0.0008354921189321393 0.0009868002859334025 -0.0014916394416246701 -8.91018
Apr 08 17:43:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      node 172_28_128_132.gra status alive monitoring/ping.go:189
Apr 08 17:43:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      skipping analyzing self node (172_28_128_131.gra) monitoring/ping.go:198
```
```
$ planet status --pretty
[...]
            {                                                                    
               "checker": "ping-checker",
               "status": "running"          
            },  
[...]
```

Seems to be working.
> * you've verified that the checker still works if nodes are removed from a cluster

```
Apr 08 17:45:11 stolonnode-1 /usr/bin/planet[11615]: DEBU [PING]      skipping node 172_28_128_133.gra because status is 'failed' monitoring/ping.go:193
```

Now there's only one remaining outstanding issue and it's that apparently:
```
tc qdisc add dev eth0 root netem delay 2000ms
```
does not affect my check O_o

Do we want to merge it anyway?

PR #98, opened 2019-02-19
allow querying custom resolvers and using NS queries

knisbet:
Updates https://github.com/gravitational/gravity/issues/257

Allow querying custom nameservers and doing NS record queries. Will be used for health checking upstream nameservers.

### DIFF
--- a/monitoring/dns.go
+++ b/monitoring/dns.go
@@ -28,7 +28,9 @@ import (
 // DNSMonitor will monitor a list of DNS servers for valid responses
 type DNSChecker struct {
 	// QuestionA is a slice of questions to ask for A (Host) records
-	QuestionA []string
+	QuestionA   []string
+	QuestionNS  []string
+	Nameservers []string
 }
 
 // Name returns the name of this checker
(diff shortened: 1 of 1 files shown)


monitoring/dns.go
```diff
@@ -28,20 +28,53 @@ import (
 // DNSMonitor will monitor a list of DNS servers for valid responses
 type DNSChecker struct {
 	// QuestionA is a slice of questions to ask for A (Host) records
-	QuestionA []string
+	QuestionA   []string
+	QuestionNS  []string
```
r0mant: Godocs.

monitoring/dns.go
```diff
@@ -55,3 +88,10 @@ func (r *DNSChecker) Check(ctx context.Context, reporter health.Reporter) {
 		Status:  pb.Probe_Running,
 	})
 }
+
+func errorDetail(question, recordType, nameserver string) string {
+	if nameserver == "" {
+		return fmt.Sprintf("failed to resolve '%v' (%v)", question, recordType)
```
r0mant: Nit: You could use `%q` instead of `'%v'`.

PR #102, opened 2019-04-10
Add pingCheck to check ping time between master nodes

eldios:
Reapplying PR (in the correct way):

https://github.com/gravitational/satellite/pull/96


monitoring/ping.go
```diff
@@ -0,0 +1,306 @@
...
+	r.Add(&pb.Probe{Checker: c.Name(), Status: pb.Probe_Running})
+}
+
+// check runs the actual system status verification code and returns an error
+// in case issues arise in the process
+func (c *pingChecker) check(ctx context.Context, r health.Reporter) (err error) {
+
+	client := c.serfClient
+
+	// retrieve other nodes using Serf members
```
a-palchikov: this is pretty obvious from the code that follows it.

monitoring/ping.go
```diff
@@ -0,0 +1,306 @@
...
+
+// checkNodesRTT implements the bulk of the logic by checking the ping time
+// between this node (self) and the other Serf Cluster member nodes
+func (c *pingChecker) checkNodesRTT(nodes []serf.Member, client agent.SerfClient) error {
+	// ping each other node and fail in case the results are over a specified
+	// threshold
+	for _, node := range nodes {
+		// skipping nodes that are not alive (failed, removed, etc..)
+		if strings.ToLower(node.Status) != strings.ToLower(pb.MemberStatus_Alive.String()) {
+			c.logger.Debugf("skipping node %s because status is '%s'", node.Name, node.Status)
```
a-palchikov: '%s' -> %q

monitoring/ping_test.go
```diff
@@ -0,0 +1,127 @@
...
+			Coords: map[string]*coordinate.Coordinate{
+				"member-1": &coordinate.Coordinate{
+					Height: 3600,
+				},
+				"member-2": &coordinate.Coordinate{
+					Height: 3600,
+				},
+			},
+			Status:      agentpb.NodeStatus_Running,
+			Description: "Testing nodes with latency value over the threshold",
```
a-palchikov: If the value is over the threshold - why is the probe running? The comment on `checkNodesRTT` explicitly states the check fails if the values are over the threshold - either the test should verify that the probe's status is `NodeStatus_Failed` or the comment rephrased to hint that this is merely a warning.

eldios: I changed the comment to specify that it just raises a warning, instead of failing.

a-palchikov: I'm just wondering why this is a warning in a log instead of a failing checker? It will hardly draw attention in this case.

agent/serf.go
```diff
@@ -131,3 +150,49 @@ type retryingClient struct {
...
+
+// MockSerfClient is a mock/fake Serf Client used in testing
+type MockSerfClient struct {
+	members []serf.Member
+	coords  map[string]*coordinate.Coordinate
+}
+
+// NewMockSerfClient is a helper function used to create a mock/fake Serf Client
+// used in testing
+func NewMockSerfClient(members []serf.Member, coords map[string]*coordinate.Coordinate) (client *MockSerfClient, err error) {
```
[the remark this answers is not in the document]
eldios: We already had a chat with Roman about this and decided that it didn't make sense to create a separated test file just for this, plus it was nice to have it here since it's logically tied to the Serf Client above (we were looking for one here initially).

If you feel that's bad let me know and I'll move it somewhere else, but we felt it was best if left in here.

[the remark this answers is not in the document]
a-palchikov: Not sure I follow but an implementation which is only useful in tests (I mean the whole `MockSerfClient` type) does not really belong here.
Besides it is not really a `mock` in the sense it is used in testing - mock objects usually record exact interactions with the API (i.e. which methods and with what parameters were invoked in a test) while this is more or a test duplicate or replacement that simply hands off dummy data.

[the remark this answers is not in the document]
eldios: okay, I'll move it to mock_serf.go then?

monitoring/ping_test.go
eldios: I initially set it that way, but @r0mant suggested this to be a warning instead.

I'll change it back to fail.

a-palchikov: No, I think it is correct at this point. This is,in fact, more like a warning which we don't model in satellite just yet (on one hand it should be visible to the user, on the other it should not be terminal to the cluster state - i.e. should not report it as `degraded`).
We do have the [Warning](https://github.com/gravitational/satellite/blob/master/agent/proto/agentpb/agent.proto#L103) severity which I would try to use in this case (i.e. `probe.State=pb.Probe_Running` and `probe.Severity = pb.Probe_Warning`. This could be handled on the gravity side by additionally outputting probes with `severity == pb.Probe_Warning`.

eldios: Added that. Please confirm it looks good and I'll (SQUASH &) merge everything. Thanks.

agent/serf.go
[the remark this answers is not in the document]
eldios: done.

PR #105, opened 2019-05-24
Add sysctl checkers that per interface forwarding is enabled

knisbet:
Updates https://github.com/gravitational/gravity.e/issues/4087

So, there's 2 potential gaps with these checks:
- I don't check that the advertise-ip interface (like ens4) has forwarding enabled. This would require detecting the name of the interface, but very weirdly, when I tried to test this on a test VM it doesn't seem like it's needed. It possible different versions/configurations of kubernetes or linux may change that, because I would expect forwarding to be needed for external traffic.
- This check can only run after the interface is created, which means we can't try and autofix these failures. A system with `net.ipv4.conf.default.forwarding = 0` will have new interfaces created without forwarding enabled, which we would have to workaround outside of satellite. Alternatively, we could have satellite check / force the default for new interfaces to forwarding = 1. 

Setting the default would be more convenient since it would just set it everywhere as we create interfaces, but I could see some customers not wanting this as a default which is why I opted for the more targeted approach.

### DIFF
--- a/monitoring/sysctl_checkers.go
+++ b/monitoring/sysctl_checkers.go
@@ -53,6 +53,50 @@ func NewMayDetachMountsChecker() *SysctlChecker {
 	}
 }
 
+// NewCNIForwardingChecker checks if CNI interface has forwarding enabled
+func NewCNIForwardingChecker() *SysctlChecker {
+	return &SysctlChecker{
+		CheckerName:     CNIForwardChecker,
+		Param:           "net.ipv4.conf.cni0.forwarding",
+		Expected:        "1",
+		OnValueMismatch: "ipv4 forwarding is off on interface cni0, see https://www.gravitational.com/docs/faq/#ipv4-forwarding",
+		SkipNotFound:    true, // interface may not exist, so skip if not found
+	}
+}
+
+// NewFlannelForwardingChecker checks if flannel interface has forwarding enabled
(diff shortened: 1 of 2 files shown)


monitoring/sysctl_checkers.go
```diff
@@ -53,11 +53,63 @@ func NewMayDetachMountsChecker() *SysctlChecker {
...
+		OnValueMismatch: "ipv4 forwarding is off on interface flannel.1, see https://www.gravitational.com/docs/faq/#ipv4-forwarding",
+		SkipNotFound:    true, // interface may not exist, so skip if not found
+	}
+}
+
+// NewWormholeBridgeForwardingChecker checks if wormhole-br0 interface has forwarding enabled
+func NewWormholeBridgeForwardingChecker() *SysctlChecker {
+	return &SysctlChecker{
+		CheckerName:     WormholeBridgeForwardChecker,
+		Param:           "net.ipv4.conf.wormhole-br0.forwarding",
```
r0mant: Are all of these interfaces always guaranteed to be present in the cluster? If any of them are missing for some reason, their corresponding checks will be failing?

a-palchikov: I don't think so - that's what I assume `SkipNotFound` is for.

r0mant: Oh lol. Sorry I looked at this before my morning coffee :man_shrugging:

PR #106, opened 2019-05-27
use third-party DNS resolver that supports root system queries

knisbet:
Updates gravitational/gravity#257

This switches the satellite DNS testing from using the default golang resolver to a more flexible external resolver. The primary reason to do this, is the embedded golang resolver doesn't consider `.` to be a valid DNS name, which prevents us from being able to run root system queries.

Due to the leaking DNS queries, we need to test the upstream DNS servers, and I think the most reliable way to do so is to perform a root system query.

### DIFF
--- a/monitoring/dns.go
+++ b/monitoring/dns.go
@@ -77,6 +77,7 @@ func (r *DNSChecker) checkWithResolver(
 			if err != nil {
 				reporter.Add(NewProbeFromErr(r.Name(), errorDetail(question, dns.TypeToString[questionType], nameserver), err))
 				checkFailed = true
+				continue
 			}
 			if in.Rcode != dns.RcodeSuccess {
 				if rcode, ok := dns.RcodeToString[in.Rcode]; ok {
--- a/monitoring/dns.go
+++ b/monitoring/dns.go
@@ -19,10 +19,12 @@ package monitoring
 import (
 	"context"
 	"fmt"
-	"net"
 
 	"github.com/gravitational/satellite/agent/health"
 	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+
+	"github.com/gravitational/trace"
+	"github.com/miekg/dns"
 )
 
 // DNSMonitor will monitor a list of DNS servers for valid responses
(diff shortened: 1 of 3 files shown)


monitoring/modules_linux_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	. "gopkg.in/check.v1"
 )
 
-type MonitoringSuite struct{}
```
knisbet: I moved this to the DNS test file that doesn't have linux build restrictions. This allows for building/running tests on non-linux platforms.

monitoring/dns.go
```diff
@@ -40,47 +42,57 @@ func (r *DNSChecker) Name() string { return "dns" }
...
+
+	for questionType, questions := range map[uint16][]string{dns.TypeA: r.QuestionA, dns.TypeNS: r.QuestionNS} {
+		for _, question := range questions {
+			q.Question[0] = dns.Question{question, questionType, dns.ClassINET}
+			in, err := dns.ExchangeContext(ctx, q, fmt.Sprintf("%s:53", nameserver))
+			if err != nil {
+				reporter.Add(NewProbeFromErr(r.Name(), errorDetail(question, dns.TypeToString[questionType], nameserver), err))
+				checkFailed = true
+			}
+			if in.Rcode != dns.RcodeSuccess {
```
a-palchikov: Will this work if `err != nil` - i.e. will `in` be consistently non-nil?

monitoring/dns_test.go
```diff
@@ -0,0 +1,84 @@
...
+				{
+					Checker: "dns",
+					Status:  pb.Probe_Running,
+				},
+			},
+			comment: "test known name (requires internet access)",
+		},
+		{
+			checker: DNSChecker{
+				QuestionA:   []string{"fdbnmfbvcnjfdblkbjcklkfldgkld.com."},
```
a-palchikov: As per https://tools.ietf.org/html/rfc6761#section-1 - have you considered using `invalid.` as the guaranteed invalid domain name?

knisbet: good point, I always forget about that one.

monitoring/dns.go
knisbet: I switched it to continue on error so it doesn't add multiple errors.

a-palchikov: In that case, I don't see why not `return` early instead of `continue`ing? Do you still want to collect failures from other queries?

knisbet: Yea, I think if multiple queries are configured, it's still relevant to run them.

PR #107, opened 2019-05-29
implement checker for file-nr

knisbet:
Updates https://github.com/gravitational/gravity.e/issues/4044

Implements a checker for open file handles vs max file handles. I set it up so it will throw a health check failure if the number of allocatable file handles falls below a minimum value. IE if the max filehandles is 100, and the minimum value is set to 10, then when more than 90 filehandles are in use the health check fill fail.

### DIFF
--- a/monitoring/sysctl.go
+++ b/monitoring/sysctl.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/gravitational/satellite/agent/health"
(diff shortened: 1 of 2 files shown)


monitoring/sysctl.go
```diff
@@ -96,6 +97,85 @@ func (c *SysctlChecker) Check(ctx context.Context, reporter health.Reporter) {
 		fmt.Sprintf("failed to query sysctl parameter %s", c.Param), trace.Wrap(err)))
 }
 
+// NewFSNRChecker creates a new checker that checks that the minimum number of file handles can be allocated
+func NewFSNRChecker(min int) health.Checker {
```
a-palchikov: `FSNR` is not too readable, `FileHandleAllocatableChecker`?

monitoring/sysctl.go
```diff
@@ -96,6 +97,85 @@ func (c *SysctlChecker) Check(ctx context.Context, reporter health.Reporter) {
...
+	}
+
+	c.check(ctx, reporter, content)
+}
+
+// check will parse the provided fileNr and validate that there are enough free file handles
+func (c *FSNRChecker) check(ctx context.Context, reporter health.Reporter, fileNr string) {
+	// parsing reference
+	// https://www.kernel.org/doc/Documentation/sysctl/fs.txt
+	split := strings.Split(fileNr, "\t")
```
a-palchikov: nit: I haven't seen in the provided reference anywhere that the values are explicitly separated with a `TAB` - why not use a more general [strings.Fields](https://golang.org/pkg/strings/#Fields)?

monitoring/sysctl.go
```diff
@@ -96,6 +97,85 @@ func (c *SysctlChecker) Check(ctx context.Context, reporter health.Reporter) {
...
+	if len(split) != 3 {
+		reporter.Add(NewProbeFromErr(FSNRCheckerID, fmt.Sprintf(
+			"fs.file-nr expected 3 fields: %v", fileNr), trace.BadParameter("expected 3 fields")))
+		return
+	}
+
+	allocatedFH, err := strconv.Atoi(split[0])
```
a-palchikov: nit: I think the suffixes only add to the confusion - it should be obvious from the context what this is about, so names like `allocated` and `max` should be self-evident.

PR #108, opened 2019-05-30
Add time drift checker.

r0mant:
This PR implements a time drift checker. The work was started by @eldios.

Note, after some consideration I've decided not to use histogram. I don't think it'd be very useful for our use-case. So time drift is compared directly with the threshold (300ms) and the check fires immediately if it's exceeded.


monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+var _ = check.Suite(&TimeDriftSuite{})
+
+func (s *TimeDriftSuite) SetUpSuite(c *check.C) {
+	s.c = &timeDriftChecker{self: node1}
+	s.clock = clockwork.NewFakeClock()
+}
+
+func (s *TimeDriftSuite) TestTimeDriftChecker(c *check.C) {
+	tests := []struct {
+		Comment string
```
a-palchikov: nit: these attributes need not be exported

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+		Nodes   []serf.Member
+		Times   map[string]time.Time
+		Result  []*pb.Probe
+	}{
+		{
+			Comment: "Time drift to both nodes is acceptable",
```
a-palchikov: "acceptable time drift"

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+func (s *TimeDriftSuite) SetUpSuite(c *check.C) {
+	s.c = &timeDriftChecker{self: node1}
+	s.clock = clockwork.NewFakeClock()
+}
+
+func (s *TimeDriftSuite) TestTimeDriftChecker(c *check.C) {
+	tests := []struct {
+		Comment string
+		Nodes   []serf.Member
+		Times   map[string]time.Time
```
a-palchikov: You could leave a comment that this is node name -> node time mapping?

agent/server.go
```diff
@@ -72,6 +74,13 @@ func (r *server) LocalStatus(ctx context.Context, req *pb.LocalStatusRequest) (r
 	return resp, nil
 }
 
+// Time sends back the target node server time
+func (r *server) Time(ctx context.Context, req *pb.TimeRequest) (*pb.TimeResponse, error) {
+	return &pb.TimeResponse{
+		Timestamp: pb.NewTimeToProto(time.Now()),
```
a-palchikov: Does not look like this is returning UTC as expected by the checker.

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+			},
+		},
+		{
+			Comment: "Time drift to both nodes is acceptable, one node is lagging behind",
+			Nodes:   []serf.Member{node1, node2, node3},
+			Times: map[string]time.Time{
+				node2.Name: s.clock.Now().Add(100 * time.Millisecond),
+				node3.Name: s.clock.Now().Add(-100 * time.Millisecond),
+			},
+			// Since we're using frozen time, latency will be 0 so
```
a-palchikov: A single comment above was enough to let in into the situation.

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+			// time drift will be exactly the amount we specified.
+			Result: []*pb.Probe{
+				s.c.successProbe(node2, 100*time.Millisecond),
+				s.c.successProbe(node3, -100*time.Millisecond),
+			},
+		},
+		{
+			Comment: "Time drift to node-3 is high",
```
a-palchikov: "time drift is over the threshold"

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+			Nodes:   []serf.Member{node1, node2, node3},
+			Times: map[string]time.Time{
+				node2.Name: s.clock.Now().Add(100 * time.Millisecond),
+				node3.Name: s.clock.Now().Add(500 * time.Millisecond),
```
a-palchikov: In this case, `500*time.Millisecond` vs `100 * time.Millisecond` does not tell me much, it would've been more readable with:
```go
node2.Name: acceptableDrift(),
node2.Name: driftOverThreshold(),
```

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+			},
+			// Since we're using frozen time, latency will be 0 so
+			// time drift will be exactly the amount we specified.
+			Result: []*pb.Probe{
+				s.c.successProbe(node2, 100*time.Millisecond),
+				s.c.failureProbe(node3, 500*time.Millisecond),
+			},
+		},
+		{
+			Comment: "Time drift to node-2 is high",
```
a-palchikov: How is this different from previous test?

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+			Nodes:   []serf.Member{node1, node2, node3},
```
a-palchikov: Also, all test cases seem to use `node1` without actually using it, so it looks out of place. I kind of get that this is the anchor node for the test, but there's a reason why good tests only specify the attributes that are tested - remove `node1` and make it an explicit anchor in the loop.

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		checker := &timeDriftChecker{
+			TimeDriftCheckerConfig: TimeDriftCheckerConfig{
+				Clock: s.clock,
+			},
+			FieldLogger: logrus.WithField(trace.Component, "test"),
+			clients:     clients,
+			serfClient:  agent.NewMockSerfClient(test.Nodes, nil),
+			name:        node1.Name,
+			self:        node1,
```
a-palchikov: That's what I'm talking about. `node1` is anchor anyways, so it does not belong in any of the test cases.

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+			// Since we're using frozen time, latency will be 0 so
+			// time drift will be exactly the amount we specified.
+			Result: []*pb.Probe{
+				s.c.failureProbe(node2, -time.Second),
+				s.c.failureProbe(node3, time.Second),
+			},
+		},
+	}
+	for _, test := range tests {
+		clients, err := ttlmap.New(10)
```
a-palchikov: Why `10`? If it's irrelevant then it only confuses the test, hide this behind an initialization function:
```go
clients := newClients(c)
// or
clients := newClients(c, test.Times, nodes) // if it depends on the test cases
```

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+				Clock: s.clock,
+			},
+			FieldLogger: logrus.WithField(trace.Component, "test"),
+			clients:     clients,
+			serfClient:  agent.NewMockSerfClient(test.Nodes, nil),
+			name:        node1.Name,
+			self:        node1,
+		}
+		var probes health.Probes
+		checker.Check(context.TODO(), &probes)
```
a-palchikov: The tests should have a tangible expiration.

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+		c.Assert(probes.GetProbes(), check.DeepEquals, test.Result,
+			check.Commentf(test.Comment))
+	}
+}
+
+type mockedTimeAgentClient struct {
+	agent.Client
```
a-palchikov: What's this for? Ah, I get it - you need the full spectrum of the client APIs. Well, in this case I'd suggest to ask only for what's needed - i.e. if you only need the timing aspect from the agent, then why not have just this instead of the full-blown client:
```go
type Timer interface {
	Time(context.Context, *pb.TimeRequest) error
}
```
and then there will be no need for any other APIs.

monitoring/timedrift_test.go
```diff
@@ -0,0 +1,189 @@
...
+		Name:   "node-2",
+		Status: agentpb.MemberStatus_Alive.String(),
+		Addr:   net.IPv4(192, 168, 0, 2),
+	}
+	node3 serf.Member = serf.Member{
+		Name:   "node-3",
+		Status: agentpb.MemberStatus_Alive.String(),
+		Addr:   net.IPv4(192, 168, 0, 3),
+	}
+	nodes map[string]serf.Member = map[string]serf.Member{
```
a-palchikov: `nodes` is not used beyond the bit that initializes `clients` - this belongs, for instance, in `newClients` instead (where the clients initialization is abstracted)

agent/proto/agentpb/agent.proto
```diff
@@ -117,7 +117,21 @@ message Timestamp {
   int32 nanoseconds = 2;
 }
 
+// TimeRequest is a request to get the node's local time
+message TimeRequest {}
+
+// TimeResponse contains the node's local time
```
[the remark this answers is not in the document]
r0mant: What is "normalized" time? If you mean timezone, it returns UTC.

monitoring/timedrift.go
```diff
@@ -0,0 +1,312 @@
...
+		return 0, trace.Wrap(err)
+	}
+
+	// Obtain this node's local timestamp.
+	//
+	// Let’s call this timestamp T1Start.
+	//
+	t1Start := c.Clock.Now().UTC()
+
+	// Send "time" request to the specified node.
```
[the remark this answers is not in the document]
r0mant: I think hiding things under single-line functions is unnecessary. The code + comments is clear as it is.

monitoring/timedrift.go
```diff
@@ -0,0 +1,312 @@
...
+
+// getAgentClient returns Satellite agent client for the provided node.
+func (c *timeDriftChecker) getAgentClient(node serf.Member) (agent.Client, error) {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+	clientI, exists := c.clients.Get(node.Addr.String())
+	if exists {
+		return clientI.(agent.Client), nil
+	}
+	client, err := c.DialRPC(&node)
```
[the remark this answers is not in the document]
r0mant: Only alive nodes are checked.

agent/server.go
r0mant: In Go timestamp always includes timezone information. But I added "UTC()" anyway.

monitoring/timedrift_test.go
r0mant: It tests negative drift.

monitoring/timedrift_test.go
r0mant: It is needed here to make sure that node does not check drift against itself.

monitoring/timedrift_test.go
r0mant: Everything is mocked and literally every single other test here does the same.

monitoring/timedrift_test.go
r0mant: TBH I never really understood/liked this approach of making a gazillion of single-method interfaces just to limit the contract for something. We're not writing a standard library, and I think in our case it only increases the complexity / ability to explore the code.

monitoring/timedrift_test.go
r0mant: Don't see anything wrong with keeping it here.

r0mant:
@a-palchikov I've addressed your comments.

agent/proto/agentpb/agent.proto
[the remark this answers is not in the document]
a-palchikov: Yes, I mean is it returning a value that can be compared with on another node. `local time` does not imply that.

monitoring/timedrift.go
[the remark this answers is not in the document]
a-palchikov: But the point of `hiding` behind single-line function is not only readability but decoupling enough and test as much as possible - is it achievable with the implementation as it is?

monitoring/timedrift.go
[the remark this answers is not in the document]
a-palchikov: How would this change if the previously alive node becomes dead?

monitoring/timedrift_test.go
a-palchikov: Ok, I also think a bogus field to satisfy a bigger interface is a superior solution.

r0mant:
@a-palchikov Yeah, I agree about the test, somehow I missed that. I'll fix it before merging.

monitoring/timedrift_test.go
r0mant: Sarcasm detected 😄I understand this is not ideal. I think I'm gonna remove this field and just provide stub implementations for the rest of methods.

PR #110, opened 2019-07-04
Bump healthz timeout.

r0mant:
To see if this alleviates issues with docker/registry/etcd healthz checks failing intermittently.

### DIFF
--- a/monitoring/healthz.go
+++ b/monitoring/healthz.go
@@ -29,7 +29,7 @@ import (
 	"github.com/gravitational/trace"
 )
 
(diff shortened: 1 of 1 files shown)


monitoring/healthz.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/gravitational/trace"
 )
 
-const healthzCheckTimeout = 1 * time.Second
+const healthzCheckTimeout = 5 * time.Second
```
a-palchikov: `10` to double the certainty?

PR #112, opened 2019-07-24
Extend node status checker to monitor node conditions.

r0mant:
This PR extends existing `node-status` checker to optionally monitor additional node conditions, in addition to `Ready` condition which is always monitored.

The primary reason is to be able to display node conditions set by [Node Problem Detector](https://github.com/kubernetes/node-problem-detector) which is now a part of Planet (see https://github.com/gravitational/planet/pull/442), but also to allow to display node conditions set by Kubernetes itself, such as `DiskPressure`, `MemoryPressure` and so on.

I have also introduced an additional probe severity, "info". A probe with "info" severity does not degrade the overall status. Right now this severity is used for conditions set by NPD because at this point I'm TBH not entirely sure it's worth degrading status for them. We can revisit this later, but this new "info" severity will also be helpful for future checks (I'm thinking firewalld and security software related).

Refs https://github.com/gravitational/gravity.e/issues/4058.


monitoring/nodesstatus.go
```diff
@@ -172,16 +230,62 @@ type kubeNodeLister struct {
...
 	// NodeStatusCheckerID identifies the checker that detects whether a node is not ready
-	NodeStatusCheckerID = "nodestatus"
+	NodeStatusCheckerID = "node-status"
 	// NodesStatusCheckerID identifies the checker that validates node availability in a cluster
-	NodesStatusCheckerID = "nodesstatus"
+	NodesStatusCheckerID = "nodes-status"
+)
+
+var (
+	// DefaultNodeConditions is a list of all default Kubernetes node
```
a-palchikov: `all` and `default` reads conflicting to me.

monitoring/nodesstatus.go
```diff
@@ -172,16 +230,62 @@ type kubeNodeLister struct {
...
+	// conditions.
+	DefaultNodeConditions = []v1.NodeConditionType{
+		v1.NodeOutOfDisk,
+		v1.NodeMemoryPressure,
+		v1.NodeDiskPressure,
+		v1.NodePIDPressure,
+		v1.NodeNetworkUnavailable,
+	}
+	// NodeProblemDetectorConditions is a list of node conditions
+	// commonly set by Node Problem Detector.
```
a-palchikov: `commonly` also has a touch of uncertainty - are they or aren't they set by NPD?
Also, since these are directly dependent on the version of planet - it should be at least mentioned to keep those in sync somehow.

agent/proto/agentpb/agent.proto
```diff
@@ -101,6 +101,9 @@ message Probe {
     // Warning defines a (possibly transient) condition that requires attention
     // but is not critical
     Warning = 2;
+    // Info defines an informational probe condition that does not
+    // lead to the degraded state
+    Info = 3;
```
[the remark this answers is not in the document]
r0mant: Could be, but warnings seem to be used everywhere now and lead to the degraded status. I don't see why this is bad though, to me it's reasonable: 

* `critical` means the problem is "very bad" and requires immediate attention and leads to degraded status. Is it actually used anywhere?
* `warning` means the problem is not as critical but cluster is still degraded.
* `info` means there is something wrong but does not lead to degradation in service.

monitoring/nodesstatus.go
```diff
@@ -89,71 +91,127 @@ func (r *nodesStatusChecker) Check(ctx context.Context, reporter health.Reporter
...
+// isNotReadyCondition returns true if the provided node condition indicates
+// that the node is not ready.
+func (r *nodeStatusChecker) isNotReadyCondition(condition v1.NodeCondition) bool {
+	return condition.Type == v1.NodeReady &&
+		condition.Status != v1.ConditionTrue
+}
+
+// isFailureCondition returns true if the provided condition is one of
+// conditions monitored by this checker and is triggered for the node.
+func (r *nodeStatusChecker) isFailureCondition(condition v1.NodeCondition) bool {
```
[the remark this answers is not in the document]
r0mant: I just tried doing that (my idea was to have the checker accept a map "condition -> expected status") by it turned out to just add more mess to the code without any benefit. The Ready condition is the only one unique in the sense that it's default value is "True" so I think it's okay that it's handled separately for now (all other conditions are initially set to False).

agent/proto/agentpb/agent.proto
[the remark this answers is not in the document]
a-palchikov: I just looked at the uses of `Probe_Warning` and only [one](https://github.com/gravitational/satellite/blob/6.0.2/monitoring/timedrift.go#L246) stood out coupled with `Probe_Failed` status as wrong to me - this should be `critical` instead. I think `warning` is precisely what you describe for `info` - a hint at some condition which does not degrade the cluster. `Info` is what we pass on as running probes.

monitoring/nodesstatus.go
```diff
@@ -89,71 +91,127 @@ func (r *nodesStatusChecker) Check(ctx context.Context, reporter health.Reporter
 	}
 }
 
+// NodeStatusCheckerConfig is the Kubernetes node status checker configuration.
+type NodeStatusCheckerConfig struct {
+	// KubeConfig provides Kubernetes access.
+	KubeConfig
+	// NodeName is the Kubernetes node name.
+	NodeName string
+	// Conditions is a list of Kubernetes node conditions to monitor.
```
a-palchikov: This still lacks context information about the type of conditions this can handle - namely, conditions whose failure is triggered by having the `status` evaluate to `v1.ConditionTrue` and not arbitrary conditions.

Also, this is somewhat limiting as a result. I can see this being useful for handling arbitrary conditions - what would it take to generalize for that? I think that something like `IsConditionTriggered(v1.NodeConditionType) bool` could work.
You would provide a custom implementation as a fall-back:
```go
func IsNPDConditionTriggered(cond v1.NodeConditionType) bool {
	return r.isNotReadyCondition(cond) || r.isFailureCondition(cond)
}
```
then any other condition could be composable as:
```go
func customIsConditionTriggered(cond v1.NodeConditionType) bool {
	return monitoring.IsNDPConditionTriggered(cond) || isAnotherConditionTriggered(cond)
}
```

monitoring/nodesstatus.go
```diff
@@ -89,71 +91,127 @@ func (r *nodesStatusChecker) Check(ctx context.Context, reporter health.Reporter
...
+	if len(nodes.Items) != 1 {
+		return nil, trace.NotFound("node %q not found", r.nodeName)
+	}
+	return &nodes.Items[0], nil
+}
+
+// isNotReadyCondition returns true if the provided node condition indicates
+// that the node is not ready.
+func (r *nodeStatusChecker) isNotReadyCondition(condition v1.NodeCondition) bool {
+	return condition.Type == v1.NodeReady &&
```
a-palchikov: Why is this a member - this does not need anything from `r`.

agent/proto/agentpb/agent.proto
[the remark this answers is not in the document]
knisbet: my 2 cents, if I were coming across this, to me I interpret as:
critical - things are on fire and something isn't working (degraded)
warning - things could potentially be on fire but are probably working. Changes to the cluster are unsafe (degraded)
info - something an admin should be aware of, but it's not clearly breaking the cluster.

Although I suspect the things we really want to report will be warnings or critical, I think I sort of agree that while we don't have a definition for certain labels that informational is appropriate (or maybe better to mark unknown since we haven't set a severity).

agent/health/health.go
```diff
@@ -99,7 +99,7 @@ func (r Probes) GetFailed() []*pb.Probe {
 func (r Probes) Status() pb.NodeStatus_Type {
 	result := pb.NodeStatus_Running
 	for _, probe := range r {
-		if probe.Status == pb.Probe_Failed {
+		if probe.Status == pb.Probe_Failed && probe.Severity != pb.Probe_Warning {
```
knisbet: Nit: not sure I'm a huge fan of writing the logic as a `!=`. If new statuses get added to the protobuf, they may interact here in an unpredictable way. Right now I believe this will cause anyone creating a probe with a Status failed and severity of None to return a status of degraded? Or any new severity to also show as degraded?

r0mant: Yeah, that's actually how it works right now (or rather, has worked until now). If you look at the triggered (failed) probes - none of them actually have severity set. I'm not sure why that is, I guess for legacy reasons, and I think we'll need to go and make sure each probe has an appropriate severity set eventually.

PR #113, opened 2019-07-29
Overlay network health testing

knisbet:
This PR does a number of things:

- Creates a new executable called nethealth, that's used to test the overlay network
- Creates a new docker image for network health testing, on quay.io/gravitational/nethealth
    - This image will be included with the monitoring-app within gravity

- There has been some interest in trying to adopt mage more widely. So I took a crack at porting the existing build targets to mage since I had to introduce a bunch of new ones anyways. It's not perfect.
  - let me know what you think. We can always roll it back, but I thought it worth trying.
- Updated some of the existing dockerfiles since I'm not sure if they're still valid or not (I had trouble building them)
  

Mage targets example:
```
➜  satellite git:(kevin/nethealth2) go run mage.go -l
Targets:
  build:all               builds all binaries
  build:buildContainer    creates a docker container as a consistent golang environment to use for software builds and tests
  build:healthz           builds the healthz binary
  build:nethealth         builds the nethealth binary
  build:satellite         builds the satellite binary
  codegen:buildbox        creates a docker container for gRPC code generation
  codegen:grpc            runs the gRPC code generator
  docker:healthz          builds the healthz container
  docker:nethealth        builds nethealth docker image
  internal:grpc           (Internal) is called from codeGen:grpc target inside docker
  publish:nethealth       tags and publishes the nethealth container to the configured registry
  test:all                runs all test targets
  test:lint               runs lints against the repo (golangci)
  test:style              validates that licenses exist
  test:unit               runs unit tests with the race detector enabled
  clean                   removes all build artifacts
```


knisbet:
While I don't believe it's guaranteed to work (any guarantee is hard), I believe ICMP echo and echo-reply are likely to be allowed within various corporate networks. Combined with while running in on-prem mode the packets are encapsulated in vxlan packets that not all firewalls / firewall rule bases will be instructed to inspect the upper layer protocols. Firewalls that are written that way though will likely give us more trouble than just icmp though, as the entire overlay network ip range would also need to be set.

So while I could envision some company creating a policy that all icmp is bad, usually that happens at the edges, as network admins get too much utility by being able to ping hosts to see if they're up. I can't say I've tried it many times, but in customer troubleshooting sessions pinging nodes or a pod often seem to work.

lib/nethealth/nethealth_test.go
```diff
@@ -0,0 +1,343 @@
...
+					lastStatusChange: server.clock.Now(),
+					addr:             &net.IPAddr{},
+				},
+			},
+			expectedLookup: map[string]string{
+				"1.1.1.1": "peer-1",
+			},
+		},
+		{
+			description: "pod that doesn't have a node set",
```
a-palchikov: Can you elaborate what exactly is expected in this test case - I don't see a difference in `expectedLookup` and would like to know where to look.

lib/nethealth/nethealth_test.go
```diff
@@ -0,0 +1,343 @@
...
+					lastStatusChange: server.clock.Now(),
+					addr:             &net.IPAddr{},
+				},
+			},
+			expectedLookup: map[string]string{
+				"1.1.1.1": "peer-1",
+			},
+		},
+		{
+			description: "pod that doesn't have an equivelant node object",
```
a-palchikov: Same here - not sure which changes (or lack thereof) to expect...

a-palchikov:
I'm having issues with the linter:
```
WARN [runner] Panic stack trace: goroutine 7037 [running]:
runtime/debug.Stack(0xe24c14, 0x12, 0xc071a30900)
	/home/travis/.gimme/versions/go1.11.5.linux.amd64/src/runtime/debug/stack.go:24 +0xa7
github.com/golangci/golangci-lint/pkg/lint.(*Runner).runLinterSafe.func1(0xc071a31d38, 0xc071a31f08)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/lint/runner.go:99 +0xb7
panic(0xca5060, 0xc0453b5320)
	/home/travis/.gimme/versions/go1.11.5.linux.amd64/src/runtime/panic.go:513 +0x1b9
github.com/golangci/golangci-lint/vendor/github.com/golangci/go-tools/lint.(*Linter).Lint(0xc071a316a8, 0xc035334b00, 0x17, 0x20, 0xc071a31670, 0x0, 0x0, 0xc0334d6980)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/vendor/github.com/golangci/go-tools/lint/lint.go:477 +0x34c2
github.com/golangci/golangci-lint/pkg/golinters.runMegacheckCheckers(0xc0385f4680, 0x4, 0x4, 0xc0397d28c0, 0xc035334b00, 0x17, 0x20, 0x4, 0x40bc1f, 0xc00c1ccb00, ...)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/golinters/megacheck.go:349 +0x1f3
github.com/golangci/golangci-lint/pkg/golinters.megacheck.runMegacheck(0x1010101, 0xc035334b00, 0x17, 0x20, 0x203000, 0x20300a, 0xc0397d2a50, 0x40bc1f, 0xc0334d6900, 0x20)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/golinters/megacheck.go:303 +0x182
github.com/golangci/golangci-lint/pkg/golinters.megacheck.Run(0xc001010101, 0xf31ea0, 0xc000255740, 0xc0277d3730, 0xc002271960, 0xdca9f7, 0x9, 0xf390e0, 0xc02901b6b0)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/golinters/megacheck.go:243 +0x111
github.com/golangci/golangci-lint/pkg/lint.(*Runner).runLinterSafe(0xc0397d2f08, 0xf31ea0, 0xc000255740, 0xc002274150, 0xc0002da380, 0x0, 0x0, 0x0, 0xf275c0, 0xc0023b6ab0)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/lint/runner.go:105 +0x161
github.com/golangci/golangci-lint/pkg/lint.Runner.runWorker.func1()
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/lint/runner.go:139 +0x64
github.com/golangci/golangci-lint/pkg/timeutils.(*Stopwatch).TrackStage(0xc02694e730, 0xdca9f7, 0x9, 0xc001bfb668)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/timeutils/stopwatch.go:75 +0x50
github.com/golangci/golangci-lint/pkg/lint.Runner.runWorker(0xc0360e5b00, 0x11, 0x11, 0xf390e0, 0xc035583110, 0xf31ea0, 0xc000255740, 0xc002274150, 0xc001d43320, 0xc001d43380, ...)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/lint/runner.go:138 +0x19c
github.com/golangci/golangci-lint/pkg/lint.(*Runner).runWorkers.func1(0xc001c3adf0, 0xc002334ae0, 0xf31ea0, 0xc000255740, 0xc002274150, 0xc001d43320, 0xc001d43380, 0xc01ec8a3c0, 0x8, 0x8, ...)
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/lint/runner.go:193 +0x13b
created by github.com/golangci/golangci-lint/pkg/lint.(*Runner).runWorkers
	/home/travis/gopath/src/github.com/golangci/golangci-lint/pkg/lint/runner.go:190 +0x1b5
```
I wonder if it the version is not stable or something.

cmd/nethealth/main.go
```diff
@@ -0,0 +1,112 @@
...
+		err = server.ListenAndServe()
+		if err != nil {
+			return trace.Wrap(err)
+		}
+
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+		sig := <-sigs
+		log.Info("Exiting on signal: ", sig)
+		return trace.Wrap(server.Stop())
```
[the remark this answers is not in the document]
knisbet: Yea my bad, apparently I pushed some half completed changes. Sorry about that.

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,587 @@
...
+		Help:      "The number of echo requests that have timed out",
+	}, []string{"node_name", "peer_name"})
+	promPeerRequest := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "nethealth",
+		Subsystem: "echo",
+		Name:      "request_total",
+		Help:      "The number of echo requests that have been sent",
+	}, []string{"node_name", "peer_name"})
+
+	prometheus.MustRegister(
```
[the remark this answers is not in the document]
knisbet: I believe MustRegister panics on failure. If you want to return and process the error there is a prometheus.Register function, which shouldn't be needed for this. Otherwise it requires placing in an init function, which is also considered a bad practice and will get noticed by the linters.

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,587 @@
...
+// and will trigger a resync if the cluster DNS service discovery changes
+func (s *Server) loopServiceDiscovery() {
+	s.Info("Starting DNS service discovery for nethealth pod.")
+	ticker := s.clock.NewTicker(dnsDiscoveryInterval)
+	query := s.config.ServiceDiscoveryQuery
+
+	previousNames := []string{}
+
+	for {
+		<-ticker.Chan()
```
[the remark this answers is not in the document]
knisbet: I don't believe that's correct. Golang routines don't block program exit.

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,587 @@
...
+		names, err := net.LookupHost(query)
+		if err != nil {
+			s.WithError(err).WithField("query", query).Error("Error querying service discovery.")
+		}
+
+		sort.Strings(names)
+		if reflect.DeepEqual(names, previousNames) {
+			continue
+		}
+		previousNames = names
```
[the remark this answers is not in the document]
knisbet: The list is sorted on line 325.

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,587 @@
...
+	for _, pod := range list.Items {
+		// skip our own pod
+		if pod.Spec.NodeName == s.config.NodeName {
+			continue
+		}
+
+		// skip if the peer object can't be located
+		if peer, ok := s.peers[pod.Spec.NodeName]; !ok {
+			continue
+		} else {
```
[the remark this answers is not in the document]
knisbet: it is required, peer is set inline, so it's only scoped to the if/else block.

knisbet:
@a-palchikov  Re: linter That's my fault. If the program has a compiler error it will fail to lint, which was caused by some half done changes which I apparently pushed. My bad they should be working with the latest push.

lib/nethealth/nethealth_test.go
knisbet: When a pod is first created, it may not have the NodeName set on the pod object. I don't expect this should occur with a daemonset, but usually it's worth validating this behavior since it's an external system. So this test simply ensures that if the kubernetes object isn't in an expected state, that it doesn't cause unexpected results in the internal structure.

lib/nethealth/nethealth_test.go
knisbet: This is a pod that is running on a node the process doesn't know about. So if we haven't collected / cached a kubernetes object, but we have a pod that points to a node we don't know about, this is meant to validate the behaviour that this is properly ignored.

lib/nethealth/nethealth.go
```diff
@@ -484,7 +482,7 @@ func (s *Server) processAck(e messageWrapper) error {
 		return nil
 	default:
 		//unexpected / unknown
-		return trace.BadParameter("received unexpected icmp message type")
+		return trace.BadParameter("received unexpected icmp message type").(trace.Error).AddField("type", e.message.Type)
```
a-palchikov: I've [updated](https://github.com/gravitational/trace/pull/50) trace type constructors to return `Error` so you don't need to type-cast:
```go
return trace.BadParameter("received unexpected icmp message type").AddField("type", e.message.Type)
```

a-palchikov:
> @a-palchikov Re: linter That's my fault. If the program has a compiler error it will fail to lint, which was caused by some half done changes which I apparently pushed. My bad they should be working with the latest push.
I don't think linter is supposed to panic on compilation errors though.

lib/nethealth/nethealth.go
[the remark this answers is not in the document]
a-palchikov: Ah, so you don't need to wait for goroutines to exit and clean up?

lib/nethealth/nethealth.go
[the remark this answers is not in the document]
a-palchikov: Yes - I somehow missed that.
On another note - is it expected to fallthrough on error from LookupHost and reset previousNames to empty and trigger resync? I guess it's nothing harmful albeit redundant.

lib/nethealth/nethealth.go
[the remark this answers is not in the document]
knisbet: Yea, I don't really see a benefit.

It's one of those things, where I had originally implemented a Stop function, but actually realized I had implemented it wrong and it was non effective. So it does tie the library code to the way it's used, because the nethealth pings can't be shutdown without process exit, but this seems like a reasonable tradeoff vs adding complexity that doesn't necessarily work right. This ties into other philosophies I'm personally a fan of like crash only software, so just exit the process as if it were to crash.

So that's why I was thinking about it a bit, and in this case I don't really see an important reason to not just let the process exit immediately, and allow the kernel to release the resources as if the process crashed.

lib/nethealth/nethealth.go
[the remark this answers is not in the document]
knisbet: It's a good point. I think this is an error since I only didn't want to process the error. I've added a `continue` to the error block.

cmd/nethealth/main.go
```diff
@@ -0,0 +1,111 @@
...
+	var (
+		app   = kingpin.New("satellite", "Cluster health monitoring agent")
+		debug = app.Flag("debug", "Enable verbose mode").Bool()
+
+		// `version` command
+		cversion = app.Command("version", "Display version")
+
+		// `run` command
+		crun               = app.Command("run", "Start nethealth agent")
+		crunPrometheusPort = crun.Flag("prom-port", "The prometheus port to bind to").Default("9801").Uint32()
```
r0mant: Is there a constant anywhere for the default port number?

cmd/nethealth/main.go
```diff
@@ -0,0 +1,111 @@
...
+		crunNamespace      = crun.Flag("namespace", "The kubernetes namespace to watch for nethealth pods").
+					Default("monitoring").OverrideDefaultFromEnvar("POD_NAMSPACE").String()
```
r0mant: `NAMSPACE` -> `NAMESPACE`

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	// heartbeatInterval is the duration between sending heartbeats to each peer. Any heartbeat that takes more
+	// than one interval to respond will also be considered timed out.
+	heartbeatInterval = 1 * time.Second
```
r0mant: Just curious, isn't this too aggressive? If nethealth implements a full mesh, there will constantly be a lot of heartbeats flying around in a large cluster?

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+
+	// Default selector to use for finding nethealth pods
+	DefaultSelector = "k8s-app=nethealth"
+
+	// DefaultServiceDiscoveryQuery is the default name to query for service discovery changes
+	DefaultServiceDiscoveryQuery = "any.nethealth"
+)
+
+const (
+	INIT    = "init"
```
r0mant: Could you add godocs on these to explain what they are? Also, a stylistic nit: why are these fully capitalized instead of `Init`, `Up` and `Timeout`?

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+	// Selector is a kubernetes selector to find all the nethealth pods in the configured namespace
+	Selector string
+	// ServiceDiscoveryQuery is a DNS name that will be used for lightweight service discovery checks. A query to
+	// any.<service>.default.svc.cluster.local will return a list of pods for the service. If the list of pods
+	// changes we know to resync with the kubernetes API. This method uses significantly less resources than running a
+	// kubernetes watcher on the API. Defaults to any.nethealth which will utilize the search path from resolv.conf.
+	ServiceDiscoveryQuery string
+}
+
+func (c Config) New() (*Server, error) {
```
r0mant: Godoc.

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+			0.0008, // 0.8 ms
+			0.0009, // 0.9 ms
+			0.001,  // 1ms
+			0.0015, // 1.5ms
+			0.002,  // 2ms
+			0.003,  // 3ms
+			0.004,  // 4ms
+			0.005,  // 5ms
+			0.01,   // 10ms
+			0.02,   // 20ms
```
r0mant: Nit: Why do 40/80ms have trailing zeros and 10/20 - don't? :)

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+			err := s.resyncPeerList()
+			if err != nil {
+				s.WithError(err).Error("Unexpected error re-syncing the list of peer nodes.")
+			}
+
+			err = s.resyncNethealthPods()
+			if err != nil {
+				s.WithError(err).Error("Unexpected error re-syncing the list of peer pods.")
+			}
+		case <-s.triggerResync:
```
r0mant: So the resync interval is 15 minutes - what happens when a node is removed from the cluster until the next resync? The other nodes will keep trying pinging it and failing? Will it trigger any alerts?

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+	for key, value := range s.addrToPeer {
+		if _, ok := s.peers[value]; !ok {
+			delete(s.addrToPeer, key)
+		}
+	}
+
+	return nil
+}
+
+// serve monitors for incomming icmp messages
```
r0mant: incomming -> incoming

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+		if err != nil {
+			log.WithError(err).Error("Error in udp socket read.")
+			continue
+		}
+
+		// The ICMP package doesn't export the protocol numbers
+		// 1 - ICMP
+		// 58 - ICMPv6
+		// https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
+		msg, err := icmp.ParseMessage(1, buf[:n])
```
r0mant: I'd make a constant for `1` with the above explanation.

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+			// Don't block
+			log.Warn("Dropped icmp message due to full rxMessage queue")
+		}
+	}
+}
+
+func (s *Server) lookupPeer(addr string) (*peer, error) {
+	peerName, ok := s.addrToPeer[addr]
+	if !ok {
+		return nil, trace.BadParameter("address not found in address table")
```
r0mant: Specify which address in the error message?

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+	}
+
+	p, ok := s.peers[peerName]
+	if !ok {
+		return nil, trace.BadParameter("peer not found in peer table")
```
r0mant: Specify which peer?

lib/nethealth/nethealth.go
```diff
@@ -0,0 +1,590 @@
...
+		Type: ipv4.ICMPTypeEcho,
+		Code: 0,
+		Body: &icmp.Echo{
+			ID:  1,
+			Seq: peer.echoCounter,
+		},
+	}
+	buf, err := msg.Marshal(nil)
+	if err != nil {
+		log.WithError(err).Warn("Failed to marshal ping.")
```
r0mant: Missing `return`? Otherwise you'll be writing to nil buffer below.

build.go
```diff
@@ -0,0 +1,306 @@
...
+
+	return trace.Wrap(sh.RunV(
+		"go", "build",
+		"-o", filepath.Join(srcDir(), "build/healthz"),
+		"-ldflags", flags(),
+		"github.com/gravitational/satellite/cmd/healthz",
+	))
+}
+
+type Publish mg.Namespace
```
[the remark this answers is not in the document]
knisbet: The namespace is sort of like a folder or group for the targets. So it helps create a sort of organization for the possible build targets. So I've grouped by build, codegen, docker, publish, test and hopefully it's a bit clears what a target does based on the grouping. 

So if you look at the possible targets, you see build related targets, publish targets, codegen, etc:
```
Targets:
  build:all               builds all binaries
  build:buildContainer    creates a docker container as a consistent golang environment to use for software builds and tests
  build:healthz           builds the healthz binary
  build:nethealth         builds the nethealth binary
  build:satellite         builds the satellite binary
  codegen:buildbox        creates a docker container for gRPC code generation
  codegen:grpc            runs the gRPC code generator
  docker:healthz          builds the healthz container
  docker:nethealth        builds nethealth docker image
  internal:grpc           (Internal) is called from codeGen:grpc target inside docker
  publish:nethealth       tags and publishes the nethealth docker container to the configured registry
  test:all                runs all test targets
  test:lint               runs lints against the repo (golangci)
  test:style              validates that licenses exist on each source file
  test:unit               runs unit tests with the race detector enabled
  clean                   removes all build artifacts
```

lib/nethealth/nethealth.go
knisbet: I don't think so. Say it's running on a 200 node cluster. Each node pings every other node, and ever node pings the local node. So that's 800 packets per second.

An icmp ping request/response is about 98 bytes on the wire, + 50 bytes in vxlan overhead. So 148 bytes * 800 = 118400 bytes/s. 947,200 bps, or still less than 1 megabit on a gigabit or 10gig link. And I suspect gravity will fall over far before hitting 200 nodes.

lib/nethealth/nethealth.go
knisbet: It depends on how it goes down. If the pod is deleted, that should trigger the service discovery change that will prevent sending any pings. However, as long as the node object exists, it will still count as down for pings. If the node object is deleted at the same time as the pod in kubernetes, it means the node will be cleaned up at that time as well within the nethealth.

However, if the pod get's deleted before the node, or it is otherwise missed somehow, then the latest this will be discovered is after 15 minutes.

This works in a similar way for when a node is added. When a node is added it doesn't take 15 minutes, the pings start as soon as coredns indicates that a new nethealth pod is available in the cluster on it's 15 second timer.

PR #117, opened 2019-09-03
Backported TimeDrift check to 5.5.x

eldios:
Backported TimeDrift check to 5.5.x

As requested by customer.


agent/client.go
```diff
@@ -126,6 +131,18 @@ func (r *client) LocalStatus(ctx context.Context) (*pb.NodeStatus, error) {
 	return resp.Status, nil
 }
 
+// Time returns the current time on the target node.
+func (r *client) Time(ctx context.Context, req *pb.TimeRequest) (time *pb.TimeResponse, err error) {
```
r0mant: Nit: I don't think named return arguments are really needed here.

agent/mock_serf.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright 2016 Gravitational, Inc.
```
r0mant: 2019

agent/client.go
eldios: makes sense, still helps with readability so I'd keep 'em. But thanks for the nit :+1:

PR #119, opened 2019-09-27
Configurable HTTP timeout

a-palchikov:
Updates https://github.com/gravitational/gravity.e/issues/4222.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -77,11 +77,11 @@ required = [
 
 [[constraint]]
   name = "github.com/prometheus/client_golang"
-  version = "v1.0.0"
+  version = "=v1.1.0"
 
 [[constraint]]
   name = "github.com/prometheus/procfs"
-  version = "v0.0.3"
+  version = "=v0.0.5"
 
 [[constraint]]
   name = "github.com/sirupsen/logrus"
--- a/build.go
+++ b/build.go
@@ -176,7 +176,7 @@ func (Test) Unit() error {
 	mg.Deps(Build.BuildContainer)
 	fmt.Println("\n=====> Running Satellite Unit Tests...\n")
 	return trace.Wrap(sh.RunV(
-		"docker", "run", "-it", "--rm=true",
+		"docker", "run", "-t", "--rm=true",
 		fmt.Sprintf("--volume=%v:/go/src/github.com/gravitational/satellite", srcDir()),
 		`--env="GOCACHE=/go/src/github.com/gravitational/satellite/build/cache/go"`,
 		`-w=/go/src/github.com/gravitational/satellite/`,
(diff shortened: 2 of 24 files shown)


Gopkg.toml
```diff
@@ -77,11 +77,11 @@ required = [
 
 [[constraint]]
   name = "github.com/prometheus/client_golang"
-  version = "v1.0.0"
+  version = "=v1.1.0"
```
a-palchikov: Did this almost subconsciously - since I tested this together with `dep update` in planet (which did bump these packages), I thought it would not hurt to update here.

build.go
```diff
@@ -176,7 +176,7 @@ func (Test) Unit() error {
 	mg.Deps(Build.BuildContainer)
 	fmt.Println("\n=====> Running Satellite Unit Tests...\n")
 	return trace.Wrap(sh.RunV(
-		"docker", "run", "-it", "--rm=true",
+		"docker", "run", "-t", "--rm=true",
```
a-palchikov: If using the `interactive` mode, terminal gets messed up if the tests/linter fail.

build.go
```diff
@@ -190,13 +190,12 @@ func (Test) Lint() error {
...
 		fmt.Sprintf("--volume=%v:/go/src/github.com/gravitational/satellite", srcDir()),
 		`--env="GOCACHE=/go/src/github.com/gravitational/satellite/build/cache/go"`,
 		fmt.Sprint("satellite-build:", version()),
-		"bash", "-c",
-		"cd /go/src/github.com/gravitational/satellite; golangci-lint run -n --deadline=30m --enable-all"+
-			" -D gochecknoglobals -D gochecknoinits",
+		"golangci-lint", "run", "--new", "--deadline=30m", "--enable-all",
```
a-palchikov: Not sure why the `bash`/`cd` was required - the Dockerfile specifies it as `WORKINGDIR`.

monitoring/etcd.go
```diff
@@ -95,6 +101,18 @@ func etcdStatus(payload []byte) (healthy bool, err error) {
 	return (result.Health == "true" || nresult.Health == true), nil
 }
 
+// NewClient returns creates a new HTTP client with default configuration
```
r0mant: either "returns" or "creates"

a-palchikov:
The `docker-registry` checker is created in planet - there will be a separate PR to use the timeout consistently.

monitoring/etcd.go
```diff
@@ -46,8 +46,14 @@ type ETCDConfig struct {
 	// InsecureSkipVerify controls whether a client verifies the
 	// server's certificate chain and host name.
 	InsecureSkipVerify bool
+	// Client specifies the optional HTTP client to use for checks.
+	// If unspecified, one will be created with default settings
+	Client *http.Client
 }
 
+// defaultHTTPTimeout defines the default HTTP client timeout for HTTP-based checks
+const defaultHTTPTimeout = 10 * time.Second
```
knisbet: More of a curiosity, any insight into why setting 10 seconds as the default or just what seems to make sense?

a-palchikov: It is more what I thought would make sense - I haven't actually measured the timeouts in those cases where it fails (for example in majority of robotests for master).

PR #124, opened 2019-10-01
Filter CONFIG_NF_NAT_IPV4 / CONFIG_NF_NAT_NEEDED on recent kernels

knisbet:
Fixes https://github.com/gravitational/gravity/issues/735

Note: I haven't setup a test VM to explicitly test this.


knisbet:
Did some testing on linux 5.1/5.2 and it seems good.

PR #127, opened 2019-10-21
Bump K8s libraries version to 1.16.2

UnderGreen:


monitoring/nodesstatus.go
```diff
@@ -243,7 +243,6 @@ var (
 	// DefaultNodeConditions is a list of default Kubernetes node
 	// conditions.
 	DefaultNodeConditions = []v1.NodeConditionType{
-		v1.NodeOutOfDisk,
```
UnderGreen: NodeOutOfDisk is removed. Use NodeDiskPressure instead.

PR #131, opened 2019-10-22
Log errors returned from serve(...)

bernardjkim:
Addresses gravitational/gravity#806

Okay to just log returned errors from `serve` or should we handle errors and shutdown satellite agent in case of error?

### DIFF
--- a/agent/server.go
+++ b/agent/server.go
@@ -30,6 +30,7 @@ import (
 	"github.com/gravitational/roundtrip"
 	"github.com/gravitational/trace"
 	serf "github.com/hashicorp/serf/client"
+	log "github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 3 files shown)


agent/server.go
```diff
@@ -126,7 +127,11 @@ func newRPCServer(agent *agent, caFile, certFile, keyFile string, rpcAddrs []str
 	handler := grpcHandlerFunc(server, healthzHandler)
 
 	for _, addr := range rpcAddrs {
-		go serve(addr, certFile, keyFile, tlsConfig, handler)
+		go func(address string) {
+			if err := serve(address, certFile, keyFile, tlsConfig, handler); err != nil {
+				log.Error(trace.DebugReport(err))
```
r0mant: I'd do `log.WithError(err).Errorf("Failed to serve on %v.", address)`.

cmd/agent/checkers.go
```diff
@@ -98,9 +98,9 @@ func addToMaster(node agent.Agent, config *config, kubeConfig monitoring.KubeCon
 	}
 
 	timeDriftHealth, err := monitoring.TimeDriftHealth(monitoring.TimeDriftCheckerConfig{
-		CAFile:     config.agentCAFile,
-		CertFile:   config.agentCertFile,
-		KeyFile:    config.agentKeyFile,
+		CAFile:     node.GetConfig().CAFile,
```
r0mant: Just curious, what was wrong with how these fields were set before?

bernardjkim: https://github.com/gravitational/satellite/blob/38ad0d30a6f68c28125b3c56a92137681f282c26/cmd/agent/main.go#L149

In cmd/agent/main.go, the `agentCAFile`, `agentCertFile`, and `agentKeyFile` are left out of the `monitorConf` variable. So the current code was not using any of those credentials when initializing the time drift checker.

We could instead just add the missing fields to `monitorConf`, but this was how planet implemented this portion of the code, so I thought it'd be best to keep it consistent.

PR #132, opened 2019-10-24
Record status/probe in case of timeout.

bernardjkim:
This PR addresses the "correct status" portion of gravitational/gravity#750 .

While collecting status, if there are any pending status queries when the context times out, they will not be included in the recorded status.
Same situation with the health checks. Any probes that timeout will not be recorded.

Instead of ignoring the node statuses and health checks that timed out, satellite should record the statuses and probes with information indicating that the query failed due to timeout.

This change will cause satellite to no longer log the warnings on timeout. If we'd still like to log the warnings after timeout, we could use a WaitGroup and a separate goroutine to check if all status queries have returned by the time the context times out.

In the situation when a network partition is created, the 'node-status' and 'kube-apiserver' checks are unable to complete and causes the entire status of the node to be dropped.

This is what `gravity status` may have returned before the change:
```
Cluster status: degraded
Cluster image:  telekube, version 7.0.0-alpha.1.122
Join token:     longtoken
Last completed operation:
    * operation_install (4d662bd6-2e4f-4bc4-b349-d16ba5b70369)
      started:          Tue Oct  8 19:09 UTC (3 hours ago)
      completed:        Tue Oct  8 19:09 UTC (3 hours ago)
Cluster endpoints:
    * Authentication gateway:
        - 172.28.128.3:32009
        - 172.28.128.4:32009
        - 172.28.128.5:32009
    * Cluster management URL:
        - https://172.28.128.3:32009
        - https://172.28.128.4:32009
        - https://172.28.128.5:32009
Cluster nodes:  dev.test
    Masters:
        * gravity1 / 172.28.128.4 / node
            Status:     degraded
        * gravity2 / 172.28.128.5 / node
            Status:     offline
    Nodes:
        * gravity0 / 172.28.128.3
            Status:     offline
...
```

This is what `gravity status` might return after the change:
```
Cluster status: degraded
Cluster image:  telekube, version 7.0.0-alpha.1.153
Join token:     longtoken
Last completed operation:
    * operation_install (1b047653-c0a6-45c0-aa8e-59b305913dff)
      started:          Wed Oct 23 22:23 UTC (1 day ago)
      completed:        Wed Oct 23 22:24 UTC (1 day ago)
Cluster endpoints:
    * Authentication gateway:
        - 172.28.128.3:32009
        - 172.28.128.5:32009
        - 172.28.128.4:32009
    * Cluster management URL:
        - https://172.28.128.3:32009
        - https://172.28.128.5:32009
        - https://172.28.128.4:32009
Cluster nodes:  dev.test
    Masters:
        * gravity0 / 172.28.128.3
            Status:     degraded
            [×]         kube-apiserver (checker timed out)
            [×]         node-status (checker timed out)
        * gravity1 / 172.28.128.4
            Status:     degraded
            [×]         kube-apiserver (checker timed out)
            [×]         node-status (checker timed out)
        * gravity2 / 172.28.128.5
            Status:     offline
...
```
The status now correctly records both nodes as 'degraded'  and all three nodes as 'master' nodes. The failed health checks are also recorded and specifies that they failed due to timeout.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -320,11 +320,15 @@ func (r *agent) runChecks(ctx context.Context) *pb.NodeStatus {
 	// channel for collecting resulting health probes
 	probeCh := make(chan health.Probes, len(r.Checkers))
 
+	ctxProbe, cancelProbe := context.WithTimeout(ctx, probeTimeout)
+	defer cancelProbe()
+
 	for _, c := range r.Checkers {
 		select {
 		case semaphoreCh <- struct{}{}:
-			go runChecker(ctx, c, probeCh, semaphoreCh)
+			go runChecker(ctxProbe, c, probeCh, semaphoreCh)
 		case <-ctx.Done():
+			cancelProbe()
 			log.Warnf("Timed out running tests: %v.", ctx.Err())
 			return emptyNodeStatus(r.name)
 		}
@@ -332,8 +336,18 @@ func (r *agent) runChecks(ctx context.Context) *pb.NodeStatus {
 
 	var probes health.Probes
 	for i := 0; i < len(r.Checkers); i++ {
-		probe := <-probeCh
-		probes = append(probes, probe...)
+		select {
+		case probe := <-probeCh:
+			probes = append(probes, probe...)
+		case <-ctx.Done():
+			cancelProbe()
+			log.Warnf("Timed out collecting test results: %v.", ctx.Err())
+			return &pb.NodeStatus{
+				Name:   r.name,
+				Status: pb.NodeStatus_Degraded,
+				Probes: probes.GetProbes(),
+			}
+		}
 	}
 
 	return &pb.NodeStatus{
@@ -364,26 +378,9 @@ func runChecker(ctx context.Context, checker health.Checker, probeCh chan<- heal
 
 	log.Debugf("Running checker %q.", checker.Name())
 
-	checkResult := make(chan health.Probes, 1)
-	go func() {
-		var probes health.Probes
-		checker.Check(ctx, &probes)
-		checkResult <- probes
-	}()
-
-	select {
-	case probes := <-checkResult:
-		probeCh <- probes
-	case <-ctx.Done():
-		var probes health.Probes
-		probes.Add(&pb.Probe{
-			Checker:  checker.Name(),
-			Status:   pb.Probe_Failed,
-			Severity: pb.Probe_Critical,
-			Error:    "checker timed out",
-		})
-		probeCh <- probes
-	}
+	var probes health.Probes
+	checker.Check(ctx, &probes)
+	probeCh <- probes
 }
 
 // statusUpdateTimeout is the amount of time to wait between status update collections.
@@ -396,6 +393,12 @@ const recycleTimeout = 10 * time.Minute
 
 // statusQueryReplyTimeout is the amount of time to wait for status query reply.
 const statusQueryReplyTimeout = 30 * time.Second
+
+// nodeTimeout specifies the amout of time to wait for a node status query reply.
+const nodeTimeout = 25 * time.Second
+
+// probeTimeout specifies the amount of time to wait for a probe to complete.
+const probeTimeout = 20 * time.Second
 
 // statusUpdateLoop is a long running background process that periodically
 // updates the health status of the cluster by querying status of other active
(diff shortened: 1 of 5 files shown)


agent/agent.go
```diff
@@ -332,17 +332,8 @@ func (r *agent) runChecks(ctx context.Context) *pb.NodeStatus {
...
-			probes = append(probes, probe...)
-		case <-ctx.Done():
-			log.Warnf("Timed out collecting test results: %v.", ctx.Err())
-			return &pb.NodeStatus{
-				Name:   r.name,
-				Status: pb.NodeStatus_Degraded,
-				Probes: probes.GetProbes(),
-			}
-		}
+		probe := <-probeCh
```
r0mant: This will potentially block indefinitely now, no?

agent/agent.go
```diff
@@ -373,9 +364,26 @@ func runChecker(ctx context.Context, checker health.Checker, probeCh chan<- heal
...
 	log.Debugf("Running checker %q.", checker.Name())
 
-	var probes health.Probes
-	checker.Check(ctx, &probes)
-	probeCh <- probes
+	checkResult := make(chan health.Probes)
+	go func() {
+		var probes health.Probes
+		checker.Check(ctx, &probes)
+		checkResult <- probes
```
r0mant: This, too, can block indefinitely and will result in a goroutine leak.

agent/agent.go
```diff
@@ -466,24 +474,16 @@ func (r *agent) collectStatus(ctx context.Context) (systemStatus *pb.SystemStatu
...
-					status.member.Name, status.member.Addr, status.err)
-				nodeStatus = unknownNodeStatus(&status.member)
-			}
-			systemStatus.Nodes = append(systemStatus.Nodes, nodeStatus)
-		case <-ctx.Done():
-			log.Warnf("Timed out collecting node statuses: %v.", ctx.Err())
-			// With insufficient status responses received, system status
-			// will be automatically degraded
-			break L
+		status := <-statusCh
```
r0mant: Here too, how is it going to timeout now if not received the status in time?

bernardjkim: I was thinking we don't want to timeout here. We should be waiting on and collecting all the node statuses. 
We are passing the context along the chain of functions required to gather status information. I thought as long as the final function call handles the timeout, all the functions higher up the chain should not block.

agent/agent.go
bernardjkim: If we can't rely on `checker.Check` to return after `ctx` is canceled, is it possible to avoid a goroutine leak or a blocked call? `Checker` has no methods to close or interrupt the on going health checks.

agent/agent.go
```diff
@@ -373,9 +364,26 @@ func runChecker(ctx context.Context, checker health.Checker, probeCh chan<- heal
 
 	log.Debugf("Running checker %q.", checker.Name())
 
-	var probes health.Probes
-	checker.Check(ctx, &probes)
-	probeCh <- probes
+	checkResult := make(chan health.Probes)
```
a-palchikov: It's safer to set chan capacity to 1 - otherwise it might block and the goroutine will be left behind.
Also it's simpler when all dependent goroutines wrap up when interrupted so we don't have to wonder if anything's stuck.

agent/agent.go
a-palchikov: Even if you don't want to exit immediately upon meeting the context deadline, it would be safer to use WaitGroup to make sure all goroutines are done before exit.
Also, this means updating `DialRPC` to accept the context as dialing a member atm is not restricted.

agent/agent.go
```diff
@@ -320,11 +321,15 @@ func (r *agent) runChecks(ctx context.Context) *pb.NodeStatus {
...
+	ctxProbe, cancelProbe := context.WithTimeout(ctx, probeTimeout)
+	defer cancelProbe()
+
 	for _, c := range r.Checkers {
 		select {
 		case semaphoreCh <- struct{}{}:
-			go runChecker(ctx, c, probeCh, semaphoreCh)
+			go runChecker(ctxProbe, c, probeCh, semaphoreCh)
 		case <-ctx.Done():
+			cancelProbe()
```
r0mant: Do you need to cancel it here and below if you deferred it above?

agent/server.go
```diff
@@ -203,7 +203,7 @@ func grpcHandlerFunc(rpcServer *server, other http.Handler) http.Handler {
 // DefaultDialRPC is a default RPC client factory function.
 // It creates a new client based on address details from the specific serf member.
 func DefaultDialRPC(caFile, certFile, keyFile string) DialRPC {
```
r0mant: Should this accept context now too?

agent/client.go
```diff
@@ -92,13 +92,18 @@ func newClient(addr, serverName, caFile, certFile, keyFile string) (*client, err
...
 // NewClientWithCreds creates a new agent RPC client to the given address
 // using specified credentials creds
-func NewClientWithCreds(addr string, creds credentials.TransportCredentials) (*client, error) {
-	conn, err := grpc.Dial(addr, grpc.WithTransportCredentials(creds))
+func NewClientWithCreds(ctx context.Context, addr string, creds credentials.TransportCredentials) (*client, error) {
+	conn, err := grpc.DialContext(
+		ctx,
+		addr,
+		grpc.WithTransportCredentials(creds),
+		grpc.WithBlock(),
```
r0mant: I'm curious, why this change to block until connection is available? I wonder if that could cause any issues with palnet-agent startup.

agent/agent.go
```diff
@@ -389,6 +395,12 @@ const recycleTimeout = 10 * time.Minute
 // statusQueryReplyTimeout is the amount of time to wait for status query reply.
 const statusQueryReplyTimeout = 30 * time.Second
 
+// nodeTimeout specifies the amout of time to wait for a node status query reply.
```
r0mant: Nit: I would explain the intent in a comment to highlight that each of the status/node/probe timeouts is supposed to be smaller than the "previous" one to allow for collection to return result before deadline. Same below.

agent/agent.go
```diff
@@ -479,6 +494,7 @@ L:
 			}
 			systemStatus.Nodes = append(systemStatus.Nodes, nodeStatus)
 		case <-ctx.Done():
+			cancelNode()
```
r0mant: Same question as above: does it need to be cancelled explicitly here?

agent/client.go
bernardjkim: I thought it would be easier to handle dial connection problems, but I guess it's not necessary.

agent/server.go
bernardjkim: I think the `DefaultDialRPC` is called when initializing the agent, but we only want to start the context when we start making `status` requests to other nodes

bernardjkim:
Still need to audit all the health probes and make sure they have timeouts set and comply with the provided context. But this patch will at least allow satellite to return partial node statuses and explicitly log that a probe may be leaking resources.

PR #135, opened 2019-11-06
Modify boot config check to comply with context

bernardjkim:
- Modify boot config checker to comply with provided context.
- Add test case to verifty checker complies with context.
- Is there a reason why some of the checkers are exported while some are not, like bootConfigParamChecker?

### DIFF
--- a/monitoring/boot_config_linux.go
+++ b/monitoring/boot_config_linux.go
@@ -80,10 +80,10 @@ func (c *bootConfigParamChecker) Check(ctx context.Context, reporter health.Repo
 	select {
 	case probes := <-probesCh:
 		health.AddFrom(reporter, probes)
-		if reporter.NumProbes() != 0 {
-			return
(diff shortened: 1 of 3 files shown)


monitoring/boot_config_linux.go
```diff
@@ -133,30 +135,35 @@ type KernelVersion struct {
 	Minor int
 }
 
-// check verifies boot configuration on host.
-func (c *bootConfigParamChecker) check(ctx context.Context, reporter health.Reporter) error {
+// check verifies boot configuration on host. Returns collected health probes.
+// Skips checks if boot configuration is not available.
+func (c *bootConfigParamChecker) check() (probes health.Reporter, err error) {
```
a-palchikov: Have `check` accept the context and update the readers to be context-aware (i.e. `parseBootConfig` would check whether the context has expired).

PR #136, opened 2019-11-06
Modify AWS health check to comply with context

bernardjkim:
- Modify AWS probe to comply with provided context.
- What test suite should aws be under? Currently added to the `MonitoringSuite`


monitoring/aws.go
```diff
@@ -38,26 +39,46 @@ type awsHasProfileChecker struct{}
...
+	probesCh := make(chan health.Reporter, 1)
+	go func() {
+		probes, err := c.check()
+		if err != nil {
+			probes.Add(NewProbeFromErr(c.Name(), "failed to validate IAM profile", err))
+		}
+		probesCh <- probes
+	}()
+	select {
+	case probes := <-probesCh:
```
a-palchikov: This is still backwards to me: the goroutine would rather handle the context and you'd wait for results (otherwise the goroutine is leaked for the duration of the request):
```go
func (c *awsHasProfileChecker) Check(ctx context.Context, reporter health.Reporter) {
	probesCh := make(chan health.Reporter, 1)
	go func() {
		probes, err := c.check(ctx)
		if err != nil {
			probes.Add(NewProbeFromErr(c.Name(), "failed to validate IAM profile", err))
		}
		probesCh <- probes
	}()
	probes := <-probesCh
	health.AddFrom(reporter, probes)
}
```
Now, it is not quite possible with the defaults (i.e. instance of EC2metadata does not use context in its APIs) but with some extra work - for example, by using a [Waiter](https://github.com/aws/aws-sdk-go/blob/v1.13.57/aws/request/waiter.go#L74):
```go
// check will check the metadata API to see if an IAM profile is assigned to the node.
func (c *awsHasProfileChecker) check(ctx context.Context) (probes health.Reporter, err error) {
	probes = &health.Probes{}

	session, err := session.NewSession()
	if err != nil {
		return probes, trace.Wrap(err, "failed to create session")
	}

	metadata := ec2metadata.New(session)
	err = c.iamInfo(ctx, metadata)
	if err != nil {
		return probes, trace.Wrap(err, "failed to determine node IAM profile")
	}

	probes.Add(NewSuccessProbe(c.Name()))
	return probes, nil
}

func (c *awsHasProfileChecker) iamInfo(ctx context.Context, metadata *ec2netadata.EC2Metadata) error {
	resp, err := c.getIAMInfo(ctx, metadata)
	if err != nil {
		return trace.Wrap(err, "failed to get EC2 IAM info")
	}

	var info ec2metadata.EC2IAMInfo
	if err := json.NewDecoder(strings.NewReader(resp)).Decode(&info); err != nil {
		return trace.Wrap(err, "failed to decode EC2 IAM info")
	}

	if info.Code != "Success" {
		return trace.Wrap(err, "failed to get EC2 IAM Info (%s)", info.Code)
	}
	return nil
}

func (c *awsHasProfileChecker) getIAMInfo(ctx context.Context, metadata *ec2metadata.EC2Metadata) error {
	waiter := &request.Waiter{
		// other fields skipped
		NewRequest: func([]request.Option) (*request.Request, error) {
			op := &request.Operation{
				Name:       "GetMetadata",
				HTTPMethod: "GET",
				HTTPPath:   path.Join("/", "meta-data", "iam/info"),
			}
			req := metadata.NewRequest(op, nil, nil)
			req.SetContext(ctx)
			return req
		},
	}
	return waiter.WaitWithContext(ctx)
}
```
The only issue with this approach, of course, is that it "leaks" some of the internal knowledge (`iam-info` endpoint).
Maybe there's a better way though...

bernardjkim:
```
func (c *awsHasProfileChecker) Check(ctx context.Context, reporter health.Reporter) {
	probesCh := make(chan health.Reporter, 1)
	go func() {
		probes, err := c.check(ctx)
		if err != nil {
			probes.Add(NewProbeFromErr(c.Name(), "failed to validate IAM profile", err))
		}
		probesCh <- probes
	}()
	probes := <-probesCh
	health.AddFrom(reporter, probes)
}
```
If we pass ctx into check then is there any need for this goroutine?
Could we just change this to:
```
func (c *awsHasProfileChecker) Check(ctx context.Context, reporter health.Reporter) {
	probes, err := c.check(ctx)
	if err != nil {
		probes.Add(NewProbeFromErr(c.Name(), "failed to validate IAM profile", err))
	}
	health.AddFrom(reporter, probes)
}
```

a-palchikov: Sure - this is even better.

monitoring/aws.go
```diff
@@ -38,26 +43,81 @@ type awsHasProfileChecker struct{}
...
+	}
+
+	if info.Code != "Success" {
+		return trace.Wrap(err, "failed to get EC2 IAM Info (%s)", info.Code)
 	}
-	reporter.Add(NewSuccessProbe(awsHasProfileCheckerID))
+	return nil
+}
+
+// getIAMInfo sends a request for IAM info and return the response content.
```
a-palchikov: and return**s** the response content.

monitoring/aws.go
```diff
@@ -38,26 +43,81 @@ type awsHasProfileChecker struct{}
...
+// Takes a context for cancelation and gets IAM info.
```
a-palchikov: ~~and gets IAM info.~~

monitoring/aws.go
```diff
@@ -38,26 +43,81 @@ type awsHasProfileChecker struct{}
...
 // Check will check the metadata API to see if an IAM profile is assigned to the node
 // Implements health.Checker
-func (*awsHasProfileChecker) Check(ctx context.Context, reporter health.Reporter) {
-	session, err := session.NewSession()
+func (c *awsHasProfileChecker) Check(ctx context.Context, reporter health.Reporter) {
+	probes, err := c.check(ctx)
 	if err != nil {
-		reporter.Add(NewProbeFromErr(awsHasProfileCheckerID, "failed to create session", trace.Wrap(err)))
-		return
+		probes.Add(NewProbeFromErr(c.Name(), "failed to validate IAM profile", err))
```
r0mant: Missing return?

monitoring/aws.go
```diff
@@ -38,26 +43,81 @@ type awsHasProfileChecker struct{}
...
+	probes.Add(NewSuccessProbe(c.Name()))
+	return probes, nil
+}
+
+// iamInfo checks if an IAM profile is assigned to the node.
+// Takes a context for cancelation.
+//
+// This is a slight modification to `EC2Metadata.IAMInfo(...)`
+// https://github.com/aws/aws-sdk-go/blob/13ad2b09f6ff7e16808b4a45c32684e6c66bf86d/aws/ec2metadata/api.go#L92
+func (c *awsHasProfileChecker) iamInfo(ctx context.Context, metadata *ec2metadata.EC2Metadata) error {
```
r0mant: Hmm, is there a better way than copying code from the go-aws SDK? I'm looking at the docs and it seems like you can provide your own HTTP client in the config, have you tried that? https://docs.aws.amazon.com/sdk-for-go/api/aws/ec2metadata/#New

a-palchikov: I came up with the following (incl. the proper `context canceled` error message so we can revert the test):
```go
// check will check the metadata API to see if an IAM profile is assigned to the node.
func (c *awsHasProfileChecker) check(ctx context.Context) (probes health.Reporter, err error) {
	probes = &health.Probes{}

	sess, err := session.NewSession()
	if err != nil {
		return probes, trace.Wrap(err, "failed to create session")
	}

	config := sess.ClientConfig(ec2metadata.ServiceName)
	config.Config.HTTPClient = &http.Client{
		Transport: newTransport(ctx),
		Timeout:   5 * time.Second,
	}
	metadata := ec2metadata.NewClient(*config.Config, config.Handlers, config.Endpoint, config.SigningRegion)
	_, err = metadata.IAMInfo()
	if isContextCanceledError(err) {
		return probes, trace.Wrap(err, "context canceled")
	}
	if err != nil {
		return probes, trace.Wrap(err, "failed to determine node IAM profile")
	}

	probes.Add(NewSuccessProbe(c.Name()))
	return probes, nil
}

func newTransport(ctx context.Context) *transport {
	return &transport{
		ctx: ctx,
		t:   http.DefaultTransport,
	}
}

func (r *transport) RoundTrip(req *http.Request) (*http.Response, error) {
	req = req.WithContext(r.ctx)
	return r.t.RoundTrip(req)
}

type transport struct {
	ctx context.Context
	t   http.RoundTripper
}

func isContextCanceledError(err error) bool {
	log.WithError(err).Info("isContextCancelledError?")
	if err == nil {
		return false
	}
	switch err := unwrapAWSError(err).(type) {
	case *url.Error:
		log.WithError(err.Err).Info("url.Error.")
		return err.Err == context.Canceled
	default:
		log.WithError(err).Info("other error.")
		return strings.Contains(err.Error(), "context canceled")
	}
}

func unwrapAWSError(err error) error {
	for err != nil {
		switch origErr := err.(type) {
		case awserr.Error:
			err = origErr.OrigErr()
		default:
			return err
		}
	}
	return err
}
```

PR #137, opened 2019-11-06
Modify cgroup checker to comply with context

bernardjkim:
- Use consistent receiver name 'c' for Checkers.
- Modify cgroup checker to comply with the provided context.
- Add test case to verify checker complies with context.

### DIFF
--- a/monitoring/cgroup.go
+++ b/monitoring/cgroup.go
@@ -116,7 +116,7 @@ func listProcMounts(ctx context.Context) ([]mountPoint, error) {
 func consistentRead(ctx context.Context, filename string, attempts int) ([]byte, error) {
 	file, err := os.Open(filename)
 	if err != nil {
-		return nil, trace.Wrap(err, "unable to open %s", filename)
+		return nil, trace.ConvertSystemError(err)
 	}
 	defer file.Close()
 
--- a/monitoring/cgroup.go
+++ b/monitoring/cgroup.go
@@ -16,9 +16,12 @@ limitations under the License.
 package monitoring
 
 import (
+	"bufio"
 	"bytes"
 	"context"
+	"io"
 	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/gravitational/satellite/agent/health"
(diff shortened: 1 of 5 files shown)


monitoring/cgroup_test.go
```diff
@@ -99,6 +100,41 @@ func (*MonitoringSuite) TestValidatesCGroupMounts(c *C) {
...
+		probes: health.Probes{
+			&pb.Probe{
+				Checker: cgroupCheckerID,
+				Detail:  "failed to validate cgroup mounts",
+				Error:   "context canceled",
+				Status:  pb.Probe_Failed,
+			},
+		},
+		reader:  testMountsReader(testCgroups),
+		comment: "contexted canceled before all checks completed",
```
a-palchikov: context~~ed~~

monitoring/cgroup.go
```diff
@@ -46,35 +46,47 @@ type cgroupChecker struct {
...
-func (r cgroupChecker) Check(ctx context.Context, reporter health.Reporter) {
-	var probes health.Probes
-	err := r.check(ctx, &probes)
-	if err != nil && !trace.IsNotFound(err) {
-		reporter.Add(NewProbeFromErr(r.Name(), "failed to validate cgroup mounts", err))
-		return
+func (c *cgroupChecker) Check(ctx context.Context, reporter health.Reporter) {
+	probesCh := make(chan health.Reporter, 1)
+	go func() {
+		probes, err := c.check()
```
a-palchikov: Here, again, the point is to cancel or at least minimize the actual work done, so shifting context inside of `check` make more sense. This needs to know how cancel the `consistentRead` operation:
 * first, it uses attempts so a basic `ctx.Done()` check inside a loop must take place
 * it uses `ioutil.ReadFile` for reading - could instead be `ioutil.ReadAll` with a reader that is context-aware.

bernardjkim:
Modified `consistentRead` to wrap the file reader with a context. `ReadWithContext`'s `Read` will return with context error if context is done. This seems to be the best we can do right now. Go does not have a simple way to interrupt a `Read` call. So we can only interrupt in between calls to `Read`?

monitoring/cgroup.go
```diff
@@ -46,35 +49,38 @@ type cgroupChecker struct {
...
-func (r cgroupChecker) Check(ctx context.Context, reporter health.Reporter) {
-	var probes health.Probes
-	err := r.check(ctx, &probes)
-	if err != nil && !trace.IsNotFound(err) {
-		reporter.Add(NewProbeFromErr(r.Name(), "failed to validate cgroup mounts", err))
-		return
+func (c *cgroupChecker) Check(ctx context.Context, reporter health.Reporter) {
+	probes, err := c.check(ctx)
+	if err != nil {
+		probes.Add(NewProbeFromErr(c.Name(), "failed to validate cgroup mounts", err))
```
r0mant: Missing return after this?

monitoring/cgroup.go
```diff
@@ -46,35 +49,38 @@ type cgroupChecker struct {
...
+func (c *cgroupChecker) check(ctx context.Context) (probes health.Reporter, err error) {
+	probes = &health.Probes{}
 
-	reporter.Add(NewSuccessProbe(r.Name()))
-}
+	mounts, err := c.getMounts(ctx)
+
+	// Skip check if mounts file is not available
+	if trace.IsNotFound(err) {
+		probes.Add(NewSuccessProbe(c.Name()))
```
r0mant: Isn't that a bit misleading that we're adding "success" probe if the check hasn't, in fact, run? As opposed to just not adding any kind of probe. Was it the same way before?

monitoring/cgroup.go
```diff
@@ -152,6 +167,25 @@ func parseProcMounts(content []byte) (mounts []mountPoint, err error) {
 	return mounts, nil
 }
 
+// ReaderWithContext wraps a reader with a context.
+type ReaderWithContext struct {
```
r0mant: Nit: This looks generic (and useful) enough to be in a some kind of "utils" package.

monitoring/cgroup.go
bernardjkim: https://github.com/gravitational/satellite/blob/de06a5d14d20373e8296f84897f624c2fdaaacb8/monitoring/cgroup.go#L53-L94
It looks like that is what we had before.
`check(...)` only adds failed probes to the reporter.
If no probes were reporter we add a success probe.

PR #138, opened 2019-11-18
Initial status timeline implementation.

bernardjkim:
Initial implementation for status timeline feature gravitational/gravity#839
- Added a history package that provides a `Timeline` interface for storing and retrieving status events.
- Added a `MemTimeline` that implements `Timeline` interface by storing the events in memory. 

TODO:
- Improve backend. Store events in a database.
- Reduce noise.
- Add a query method to the `Timeline` interface to query events based on event labels or values.
- ...


agent/proto/agentpb/agent.proto
```diff
@@ -126,6 +126,14 @@ message TimeResponse {
   Timestamp timestamp = 1;
 }
 
+// TimelineRequest is a request to get the cluster status timeline
+message TimelineRequest {}
+
+// TimelineResponse contains the cluster status timeline events
+message TimelineResponse {
+  repeated string timeline = 1;
```
r0mant: I think we should return a structured response, as a list of timeline events for example.

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// TODO: can we rely on user CLI sessions to provide colored text?
+	// If not, maybe we should indicate the different event types a different
+	// way.
+
+	// Color specifies the color of the event.
+	// Red -> Indicates change has caused state to degraded.
+	// Yellow -> Indicates change has not caused any state changes.
+	// Green -> Indicates changes has caused state to recover.
+	Color string
```
r0mant: Can we reuse existing probe statuses instead of introducing another thing to basically represent the same thing? Then the caller can output them appropriately (provided we return structured response, see comment above).

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+func (e *Event) String() string {
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("[%s] ", e.TimeStamp.Format(Stamp)))
+	sb.WriteString(e.Description)
+	return sb.String()
+}
+
+// Timeline represents a timeline of cluster status events. The Timeline
+// can hold a specified amount of events and uses a FIFO eviction policy.
+type Timeline struct {
```
r0mant: Suggestion: I would make Timeline an interface so it's easy to substitute implementations. For example, this first version of timeline keeps stuff in memory, afterwards we'll add a "prometheus" or "sqlite" backend as discussed before. It will be also helpful in tests which could provide their own implementation.

agent/agent.go
```diff
@@ -135,6 +135,7 @@ func New(config *Config) (Agent, error) {
 	agent := &agent{
 		Config:          *config,
 		SerfClient:      client,
+		Timeline:        timeline.NewTimeline(256),
```
r0mant: I would pass the size via Config struct.

agent/server.go
```diff
@@ -82,6 +82,14 @@ func (r *server) Time(ctx context.Context, req *pb.TimeRequest) (*pb.TimeRespons
 	}, nil
 }
 
+// Timeline sends the current status timeline.
+func (r *server) Timeline(ctx context.Context, req *pb.TimelineRequest) (*pb.TimelineResponse, error) {
+	timeline := r.agent.Timeline.GetTimeline()
+	resp := &pb.TimelineResponse{}
+	resp.Timeline = timeline
```
r0mant:
```go
resp := &pb.TimelineResponse{
  Timeline: timeline,
}
```

then?

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+)
+
+// Stamp defines default timestamp format for events.
+const Stamp = "Jan _2 15:04:05"
+
+// Event represents a Timeline event. An event occurrs  whenever there is a
+// change in the cluster status. An event could be triggered by a failed
+// heartbeat or a failed health check.
```
r0mant: Technically, a failed probe returning back to a healthy status also triggers event, the comment should probably reflect this.

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+
+// Timeline represents a timeline of cluster status events. The Timeline
+// can hold a specified amount of events and uses a FIFO eviction policy.
+type Timeline struct {
+	// Size specifies the max size of the timeline.
+	Size int
+	// Timeline holds the latest status events.
+	Timeline []string
+	// Cluster holds the latest cluster status.
+	Cluster *Cluster
```
r0mant: Nit: I'd name this `LastStatus` or something.

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// NewTimeline initializes and returns a new StatusTimeline with the
+// specified size.
+func NewTimeline(size int) Timeline {
+	return Timeline{
+		Size:     size,
+		Timeline: []string{},
```
r0mant: `make([]string, 0, size)`?

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+	return sb.String()
+}
+
+// Timeline represents a timeline of cluster status events. The Timeline
+// can hold a specified amount of events and uses a FIFO eviction policy.
+type Timeline struct {
+	// Size specifies the max size of the timeline.
+	Size int
+	// Timeline holds the latest status events.
+	Timeline []string
```
r0mant: I think this should also be something structured instead of just strings. I.e. I would define an `Event` gRPC object that would be returned in timeline via gRPC API and use that here as well.

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+
+// RecordStatus records differences of the previous status to the provided
+// status into the Timeline.
+func (t *Timeline) RecordStatus(status *pb.SystemStatus) {
+	cluster := parseSystemStatus(status)
+	events := t.Cluster.diffCluster(cluster)
+	for _, event := range events {
+		t.addEvent(event)
+	}
+	t.Cluster = cluster
```
r0mant: This (as well as this method in general and other methods that access shared state) should be protected by a mutex.

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+
+// Cluster represents the overall status of a cluster.
+type Cluster struct {
+	// Status specifies the cluster status.
+	Status string
+	// Nodes specify the individual node statuses.
+	Nodes map[string]*Node
+}
+
+// difCluster calculates the differences from the provided cluster and returns
```
r0mant: diff

lib/timeline/timeline.go
```diff
@@ -0,0 +1,280 @@
...
+// the differences as a list of events.
+func (c *Cluster) diffCluster(cluster *Cluster) []Event {
+	events := []Event{}
+
+	// Compare cluster status
+	if c.Status != cluster.Status {
+		event := NewEvent()
+		event.Description = fmt.Sprintf("cluster status changed from [%s] to [%s]", c.Status, cluster.Status)
```
r0mant: If I was doing this, I would try to decouple the representation of the objects (i.e. descriptions) from the objects themselves. For example, we could have an event object:

```go
type TimelineEvent struct {
  Type string // defines event type, e.g. "status", "nodeAdded", "nodeRemoved", "probeFailed", etc.
  Metadata map[string]string // event-specific metadata e.g. node address, old/new status, etc.
}
```

Then these events can be passed on the wire over gRPC API and the callers can format them appropriately based on the event type and its metadata.

lib/history/timeline.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Keep track of removed probes
+	removed := map[string]bool{}
+	for name := range n.Probes {
+		removed[name] = true
+	}
+
+	// Probes added or modified
+	for name, newProbe := range node.Probes {
+		if oldProbe, ok := n.Probes[name]; !ok {
```
a-palchikov: Handle the happy (`ok == true`) case first and continue - w/o the need for `else`.

vagrant/ansible/ssh-config
```diff
@@ -0,0 +1,33 @@
+Host k8s-master
+  HostName 127.0.0.1
+  User vagrant
+  Port 2222
+  UserKnownHostsFile /dev/null
+  StrictHostKeyChecking no
+  PasswordAuthentication no
+  IdentityFile /home/bernard/.vagrant.d/insecure_private_key
```
a-palchikov: I think this should be configurable (also see below)

lib/history/event.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Event represents a single timeline event. An event exposes a type and
+// metadata.
+type Event struct {
+	// timeStamp specifies the when the event occurred.
+	timeStamp time.Time
+	// eventType specifies the type of event.
+	eventType EventType
+	// mu locks access to metadata
+	mu sync.Mutex
```
a-palchikov: I don't think the events need a mutex - the metadata can be passed to an event during construction.

vagrant/ansible/ssh-config
bernardjkim: oops. I think ssh-config and inventory will be auto generated, so I'll just rm from repo

lib/history/timeline.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// Compare cluster status
+	if c.Status != cluster.Status {
+		var event *Event
+		if cluster.Status == pb.SystemStatus_Running.String() {
+			event = NewClusterRecoveredEvent()
+		} else {
+			event = NewClusterDegradedEvent()
+		}
+		event.SetMetadata("old", c.Status)
```
a-palchikov: I would even unexport `SetMetadata` so events are values to the outside.

bernardjkim:
Going to go ahead and merge PR for initial timeline implementation. Going to continue working on timeline feature in branch `bernard/timeline` and create separate PRs to that branch for additional TODOs.

PR #141, opened 2019-12-11
Timeline consistency

bernardjkim:
This PR implements timeline consistency portion of the timeline feature gravitational/gravity#839.

Implementation design doc can be found [here](https://docs.google.com/document/d/1sM33QHLvf3-zbNfvcP5R_2lbzeGiLeQhKFGc730on0A/edit#)

Each satellite agent now only makes updates in the scope of that node. This should only include changes to the node's status and its health probes.

In order for timeline to be replicated within a cluster, nodes will need to request timelines from another node and merge the requested timeline into their current timeline.
We can explore the other methods of replicating the timeline if we need to improve efficiency.

In order to record events outside the scope of a node (cluster status, cluster upgrade, cluster resize, etc..), gravity/planet will need to have the leader node track these events and either store their own timeline, or send the events to the satellite agent to be recorded. `UpdateTimeline` RPC has been provided for this purpose.

TODO: integration.
TODO: reinitialize previous status in case satellite is restarted
TODO: indexing for faster query


agent/proto/agentpb/agent.proto
```diff
@@ -172,4 +181,7 @@ service Agent {
   rpc Time(TimeRequest) returns (TimeResponse) {}
   // Timeline collects the status timeline
   rpc Timeline(TimelineRequest) returns (TimelineResponse) {}
+  // UpdateTimeline requests a new event be added to the timeline.
+  // This RPC is idempotent. Repeated calls will have no effect.
```
a-palchikov: Repeated as in with the same event value?

lib/history/memory/inmemory_test.go
```diff
@@ -43,32 +43,43 @@ func (s *InMemorySuite) SetUpSuite(c *C) {
...
 	var timeline history.Timeline
 	timeline = NewTimeline(s.clock, 1)
-	err := timeline.RecordStatus(context.TODO(), &pb.SystemStatus{Status: pb.SystemStatus_Running})
+
+	node := "test-node"
+	old := &pb.NodeStatus{Name: node, Status: pb.NodeStatus_Running}
+	new := &pb.NodeStatus{Name: node, Status: pb.NodeStatus_Degraded}
+
+	var err error
+	err = timeline.RecordStatus(context.TODO(), old)
```
a-palchikov: I would allot a specific time for the test.

lib/history/sqlite/constants.go
```diff
@@ -0,0 +1,59 @@
+package sqlite
+
+import "time"
+
+const (
+	// evictionFrequency is the time between eviction loops.
+	evictionFrequency = time.Hour
+
+	// evictionTimeout specifies the amount of time to given to evict events.
```
a-palchikov: ~~to~~ given

lib/history/sqlite/event.go
```diff
@@ -0,0 +1,98 @@
...
+	Node      sql.NullString `db:"node"`
+	Probe     sql.NullString `db:"probe"`
+	Old       sql.NullString `db:"oldState"`
+	New       sql.NullString `db:"newState"`
+}
+
+// newSQLEvent constructs a new sqlEvent from the provided TimelineEvent.
+func newSQLEvent(event *pb.TimelineEvent) (row sqlEvent, err error) {
+	row = sqlEvent{Timestamp: event.GetTimestamp().ToTime()}
+	switch t := event.GetData().(type) {
```
a-palchikov: This actually lends itself perfectly to interfaces - instead of exhaustively enumerating all known types (which would require changes with each new type added) factor this into an interface for events to implement and avoid the bogus `trace.BadParameter("unknown event type %T", t)`

lib/history/sqlite/event.go
```diff
@@ -0,0 +1,98 @@
...
+		row.Node = sql.NullString{String: e.GetNode(), Valid: true}
+		row.Probe = sql.NullString{String: e.GetProbe(), Valid: true}
+	default:
+		return row, trace.BadParameter("unknown event type %T", t)
+	}
+	return row, nil
+}
+
+// toArgs returns sqlEvent as a list of arguments.
+func (e *sqlEvent) toArgs() (args []interface{}) {
```
a-palchikov: same here - invert the direction since it does not tell me anything about where and for each purpose the list is. Instead, use a descriptive and focused interface (like `ValueInserter`) which would accept the DB instance.

lib/history/sqlite/event.go
```diff
@@ -0,0 +1,98 @@
...
+		e.Node.String,
+		e.Probe.String,
+		e.New.String,
+		e.Old.String,
+	}
+}
+
+// toProto returns sqlEvent as a protobuf TimelineEvent.
+func (e *sqlEvent) toProto() (*pb.TimelineEvent, error) {
+	switch e.EventType {
```
a-palchikov: Sounds like another case for the interface.

lib/history/sqlite/sqlite_test.go
```diff
@@ -63,48 +65,68 @@ func (s *SQLiteSuite) TearDownTest(c *C) {
...
+	err = s.timeline.evictEvents(context.TODO(), s.clock.Now())
 	c.Assert(err, IsNil)
 
 	actual, err := s.timeline.GetEvents(context.TODO(), nil)
 	c.Assert(err, IsNil)
 
-	expected := []*pb.TimelineEvent{history.NewClusterDegraded(s.clock.Now())}
-	c.Assert(actual, DeepEquals, expected, Commentf("Test FIFO eviction"))
+	for _, event := range actual {
+		fmt.Println(event)
```
a-palchikov: That should be optional.

lib/history/sqlite/sqlite_test.go
```diff
@@ -63,48 +65,68 @@ func (s *SQLiteSuite) TearDownTest(c *C) {
...
+
+func (s *SQLiteSuite) TestMergeEvents(c *C) {
+	events := []*pb.TimelineEvent{history.NewNodeDegraded(s.clock.Now(), "test-node")}
+	err := s.timeline.RecordTimeline(context.TODO(), events)
+	c.Assert(err, IsNil)
+
+	actual, err := s.timeline.GetEvents(context.TODO(), nil)
+	c.Assert(err, IsNil)
+
+	c.Assert(actual, DeepEquals, events, Commentf("Test merge events"))
```
a-palchikov: nit: the comment should reflect what was expected to clarify if the test fails.

lib/history/sqlite/utils.go
```diff
@@ -0,0 +1,15 @@
+package sqlite
+
+import (
+	"github.com/mattn/go-sqlite3"
+)
+
+// isErrConstraintUnique returns true if error is a
+// sqlite3.ErrorConstraintUnique error.
+func isErrConstraintUnique(err error) bool {
+	sqliteErr, ok := err.(sqlite3.Error)
```
a-palchikov: Make sure the specified `err` has not been wrapped with `trace.Wrap` or do s safe `trace.Unwrap` to make sure you're always dealing with the original error.

lib/history/sqlite/sqlite.go
```diff
@@ -57,88 +53,119 @@ type Timeline struct {
...
-// getSize returns the current number of rows stored in the provided database.
-func getSize(ctx context.Context, database *sqlx.DB) (size int, err error) {
-	row := database.QueryRowContext(ctx, "SELECT COUNT(*) FROM events")
-	if err := row.Scan(&size); err != nil {
-		return -1, trace.Wrap(err)
+// eventEvictionLoop periodically evicts old events to free up storage.
+func (t *Timeline) eventEvictionLoop() {
+	var timer <-chan time.Time
+	for {
+		timer = t.config.Clock.After(evictionFrequency)
```
knisbet: Nit: this looks like it would be better suited to be a ticker, which is supported by ClockWork. So something like:

```
ticker := t.config.Clock.NewTicker(evictionFrequency)
for {
                  <-ticker.Chan()
    
                 // All events before this time will be deleted.
		retentionCutOff := t.config.Clock.Now().Add(-(t.config.RetentionDuration))

		ctx, cancel := context.WithTimeout(context.Background(), evictionTimeout)
		if err := t.evictEvents(ctx, retentionCutOff); err != nil {
			cancel()
			log.WithError(err).Warnf("Error evicting expired events.")
		}
}
```

Not a huge deal in this case, just saves a line, and a tiny memory allocation each loop.

agent/proto/agentpb/agent.proto
bernardjkim: Yes, will specify.

agent/proto/agentpb/agent.proto
```diff
@@ -172,4 +181,7 @@ service Agent {
...
+  rpc UpdateTimeline(UpdateRequest) returns (UpdateResponse) {}
```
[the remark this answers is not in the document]
bernardjkim: With the current implementation, the satellite/planet agents will only be keeping track of events in the scope of that node, so this only includes the node status and health check results.
In order to keep track of events such as gravity upgrade, cluster resize, ... these events will only be visible by gravity?
So this RPC is to be used by the currently elected gravity leader and send these types of events to satellite.
You mentioned that gravity provides an audit log. I haven't yet looked into exactly what data is provided. This RPC is currently not being used, and may not be needed.

lib/history/sqlite/event.go
bernardjkim: Still a little confused on how to structure the interfaces. Are you looking for something like this?

```golang
// SQLInserter inserts row into underlying SQL database.
type SQLInserter interface {
  // InsertRow prepares row to be inserted.
  InsertRow(SQLRow) error
  // ExecInsert inserts all prepared rows to underlying storage.
  ExecInsert() error
}

// SQLRow can be inserted into a SQL database.
type SQLRow interface {
  // Insert inserts row into storage.
  Insert(*sql.DB) error
}

// ProtoBuffer can be converted into a protobuf TimelineEvent.
type ProtoBuffer interface {
  // ProtoBuf returns event as a protobuf message.
  ProtoBuf() *pb.TimelineEvent, error
}
```
SQLInserter can be implemented with different sql databases.
We can then define separate structs for each event type and each event type implements `SQLRow` and `ProtoBuffer` to define their own `Insert` and `ProtoBuf` methods.

lib/history/sqlite/constants.go
```diff
@@ -0,0 +1,59 @@
...
+	type,
+	node,
+	probe,
+	oldState,
+	newState
+) VALUES (?,?,?,?,?,?)
+`
+
+// deleteOldFromEvents is sql statement to delete entries from `events` table.
+const deleteOldFromEvents = `DELETE FROM events WHERE id IN (SELECT id FROM events WHERE timestamp < ?)`
```
r0mant: Wouldn't just `DELETE FROM events WHERE timestamp < ?` work? The DELETE clause supports WHERE according to https://www.sqlite.org/lang_delete.html.

lib/history/sqlite/event.go
```diff
@@ -0,0 +1,329 @@
+package sqlite
```
r0mant: Missing license header.

lib/history/sqlite/event.go
```diff
@@ -0,0 +1,329 @@
...
+	ID int `db:"id"`
+	// Timestamp specifies event timestamp.
+	Timestamp time.Time `db:"timestamp"`
+	// EventType specifies event type.
+	EventType string `db:"type"`
+	// Node specifies name of node.
+	Node sql.NullString `db:"node"`
+	// Probe specifies name of probe.
+	Probe sql.NullString `db:"probe"`
+	// Old specifies previous value.
```
r0mant: Nit: "previous probe state" maybe? Same below.

lib/history/sqlite/sqlite.go
```diff
@@ -30,115 +29,142 @@ import (
...
-		size:     size,
-		database: database,
-		// TODO: store and recover lastStatus in case satellite agent restarts.
-		lastStatus: nil,
-	}, nil
+	// if err := timeline.initPrevStatus(ctx); err != nil {
+	// 	return nil, trace.Wrap(err, "failed to init previous status")
+	// }
+
+	// TODO: should this be called in constructor method?
```
r0mant: I think it's fine unless the caller can have a reason to instantiate the timeline multiple times?

lib/history/sqlite/sqlite.go
```diff
@@ -30,115 +29,142 @@ import (
...
 }
 
-// getSize returns the current number of rows stored in the provided database.
-func getSize(ctx context.Context, database *sqlx.DB) (size int, err error) {
-	row := database.QueryRowContext(ctx, "SELECT COUNT(*) FROM events")
-	if err := row.Scan(&size); err != nil {
-		return -1, trace.Wrap(err)
+// eventEvictionLoop periodically evicts old events to free up storage.
+func (t *Timeline) eventEvictionLoop() {
+	ticker := t.config.Clock.NewTicker(evictionFrequency)
```
r0mant: `defer ticker.Stop()`

lib/history/sqlite/sqlite.go
```diff
@@ -30,115 +29,142 @@ import (
...
+	for {
+		<-ticker.Chan()
```
r0mant: So I would rewrite this method a little bit and have it accept an external context and use select statement so the caller can potentially cancel it. Something like:

```go
func (t *Timeline) eventEvictionLoop(ctx context.Context) {
  ticker := ...
  defer ticker.Stop()
  for {
    select {
      <-ticker.C:
        localCtx, cancel := context.WithTimeout(ctx, evictionTimeout)
        // call t.evictEvents(localCtx, ...)
      <-ctx.Done():
        log.Info("Eviction loop is stopping.")
        return
    }
  }
}        
```

lib/history/sqlite/sqlite.go
```diff
@@ -30,115 +29,142 @@ import (
...
+
+// RecordTimeline merges the provided events into the current timeline.
+// Duplicate events will be ignored.
+func (t *Timeline) RecordTimeline(ctx context.Context, events []*pb.TimelineEvent) (err error) {
+	if len(events) == 0 {
+		return nil
+	}
+
+	// All events before this time will be deleted.
+	retentionCutOff := t.config.Clock.Now().Add(-(t.config.RetentionDuration))
```
r0mant: Nit: I notice this incantation is used here for the second time, I would consider moving it out into a single-line helper method, like `func getRetentionCutoff() time.Time` or something.

lib/history/sqlite/sqlite.go
```diff
@@ -30,115 +29,142 @@ import (
...
-		t.mu.Unlock()
+		t.statusMutex.Unlock()
 		return nil
 	}
-
-	t.size = t.size + len(events)
-	if t.size > t.Config.Capacity {
-		t.size = t.Config.Capacity
-	}
 	t.lastStatus = status
```
r0mant: Nit: I usually try to add a bit more debugging information to ease troubleshooting later on or just be able to see what's going on, so for example if a status changes, I would log the events or something.

lib/history/sqlite/sqlite.go
```diff
@@ -30,115 +29,142 @@ import (
...
+		return nil
+	}
+
+	// All events before this time will be deleted.
+	retentionCutOff := t.config.Clock.Now().Add(-(t.config.RetentionDuration))
+
+	// Filter out expired events.
+	filtered := []*pb.TimelineEvent{}
+	for _, event := range events {
+		if event.GetTimestamp().ToTime().After(retentionCutOff) {
```
r0mant: Here too: I would log the events that are filtered out, if only just for informational purposes. There can't be too much debug information :) (not true, I know...).

On another note - is it necessary to filter these out here on the receiving side? I think that either the client can filter them before sending (which will arguably save a bit of traffic too), or just insert everything and let the regular gc loop clean them up on the next iteration?

bernardjkim: In the current implementation, the gc cycle runs at a much lower frequency than the update cycle. We can't expect the nodes to have synchronized gc cycles either. So if we don't filter out the events when they are received, expired events may continually circulate through the cluster.

We should look to improve the implementation and reduce network traffic. One way would be to broadcast timeline updates instead of merging timelines.
Will work on optimization here a little later down the line.

lib/history/sqlite/sqlite.go
bernardjkim: The eviction loop is called from the timeline constructor method. In this situation should the constructor take an additional context parameter? Should the timeline struct initialize its own context to be used for the eviction loop and have some kind of shutdown method to cancel the context?

lib/history/memory/inmemory_test.go
```diff
@@ -41,34 +42,54 @@ func (s *InMemorySuite) SetUpSuite(c *C) {
 }
 
 func (s *InMemorySuite) TestRecordStatus(c *C) {
+	ctx, cancel := context.WithTimeout(context.TODO(), testTimeout)
+	defer cancel()
+
 	var timeline history.Timeline
```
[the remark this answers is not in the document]
bernardjkim: The timeline had the interface type so that it would test that the struct implemented the interface. I guess it'd be better to test that somewhere else.

lib/history/memory/inmemory_test.go
```diff
@@ -41,34 +42,54 @@ func (s *InMemorySuite) SetUpSuite(c *C) {
...
+	err = timeline.RecordStatus(ctx, new)
 	c.Assert(err, IsNil)
 
-	actual, err := timeline.GetEvents(context.TODO(), nil)
+	actual, err := timeline.GetEvents(ctx, nil)
 	c.Assert(err, IsNil)
 
-	expected := []*pb.TimelineEvent{history.NewClusterRecovered(s.clock.Now())}
+	expected := []*pb.TimelineEvent{history.NewNodeDegraded(s.clock.Now(), node)}
 	c.Assert(actual, DeepEquals, expected, Commentf("Test record status"))
```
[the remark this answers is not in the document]
bernardjkim: Will clean up tests and add some comments to improve readability.

PR #142, opened 2019-12-12
Timeline integration

bernardjkim:
Integrate timeline into satellite. 

- Satellite agent now takes additional `DBPath` and `Retention` to configure timeline.
- Satellite agent runs a `timelineUpdateLoop` that periodically updates the timeline with local node changes and merges the timeline from one other random node.
- Satellite agent also provides a `UpdateTimeline` RPC to be used to update the timeline.

- In the current implementation, each node pulls timelines from each other node in the cluster and merges with their own view of the timeline. This will guarantee that the cluster will have an eventual consistent view of the status timeline. This design can be improved. 

- Instead of constantly merging timelines, we can push updates when available. This will reduce unnecessary traffic. 

- We can further optimize by having the overall timeline only be recorded by master nodes.


agent/server_test.go
```diff
@@ -569,16 +567,17 @@ func (r *AgentSuite) newAgent(node string, rpcPort int, members []serf.Member,
...
+		name:          node,
+		SerfClient:    &MockSerfClient{members: members},
+		dialRPC:       testDialRPC(rpcPort, r.certFile, r.keyFile),
+		cache:         &testCache{c: c, SystemStatus: &pb.SystemStatus{Status: pb.SystemStatus_Unknown}},
+		Checkers:      checkers,
+		statusClock:   statusClock,
+		recycleClock:  recycleClock,
+		timelineClock: clockwork.NewFakeClock(),
+		localStatus:   emptyNodeStatus(node),
+		Timeline:      nil,
```
a-palchikov: nit: redundant.

cmd/agent/main.go
```diff
@@ -74,9 +75,13 @@ func run() error {
 		cagentCAFile         = cagent.Flag("ca-file", "SSL CA certificate for verifying server certificates").ExistingFile()
 		cagentCertFile       = cagent.Flag("cert-file", "SSL certificate for server RPC").ExistingFile()
 		cagentKeyFile        = cagent.Flag("key-file", "SSL certificate key for server RPC").ExistingFile()
+		// sqlite config
+		cagentDBPath    = cagent.Flag("db-path", "SQLite database location").Default("/tmp/timeline.db").String()
+		cagentRetention = cagent.Flag("retention", "Timeline retention duration").Duration()
```
a-palchikov: `Time window to retain the timeline for as Go duration.`

cmd/agent/main.go
```diff
@@ -74,9 +75,13 @@ func run() error {
...
 
 		// `status` command
 		cstatus            = app.Command("status", "Query cluster status")
+		cstatusHistory     = cstatus.Flag("history", "Query the status history").Bool()
```
a-palchikov: Why not a sub-command?

cmd/agent/main.go
```diff
@@ -135,16 +140,24 @@ func run() error {
 			}
 			backends = append(backends, influxdb)
 		}
+
+		// Use retention duration of 1 week if no duration is provided
+		if *cagentRetention == time.Duration(0) {
+			*cagentRetention = time.Hour * 24 * 7
```
a-palchikov: move to a const with a descriptive name

cmd/agent/status.go
```diff
@@ -38,7 +38,7 @@ const statusTimeout = 5 * time.Second
 // the status of the cluster.
 // Returns true if the status query was successful, false - otherwise.
 // The output is prettified if prettyPrint is true.
-func status(RPCPort int, local, prettyPrint bool, caFile, certFile, keyFile string) (ok bool, err error) {
+func status(RPCPort int, history, local, prettyPrint bool, caFile, certFile, keyFile string) (ok bool, err error) {
```
a-palchikov: I think it's time for `statusConfig`

cmd/agent/status.go
```diff
@@ -48,7 +48,20 @@ func status(RPCPort int, local, prettyPrint bool, caFile, certFile, keyFile stri
 	if err != nil {
 		return false, trace.Wrap(err)
 	}
-	var statusJson []byte
+
+	if history {
```
a-palchikov: Sounds like a separate command to me.

cmd/agent/status.go
```diff
@@ -66,15 +79,47 @@ func status(RPCPort int, local, prettyPrint bool, caFile, certFile, keyFile stri
...
 	return ok, nil
 }
+
+func printEvent(event *pb.TimelineEvent) {
+	timestamp := event.GetTimestamp().ToTime()
+	fmt.Printf("[%s] ", timestamp.Format(Stamp))
+
+	switch event.GetData().(type) {
+	case *pb.TimelineEvent_ClusterDegraded:
+		fmt.Printf("Cluster Degraded\n")
```
a-palchikov: `fmt.Println` without the trailing `\n`

agent/agent.go
```diff
@@ -134,20 +138,33 @@ func New(config *Config) (Agent, error) {
 		return nil, trace.Wrap(err, "failed to serve prometheus metrics")
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), timelineInitTimeout)
+	defer cancel()
+
+	timeline, err := sqlite.NewTimeline(ctx, sqlite.Config{
+		DBPath:            config.DBPath,
+		RetentionDuration: config.RetentionDuration,
+		Clock:             clockwork.NewRealClock(),
```
r0mant: Nit: The way we usually do it is for things like that we don't pass them explicitly but have a CheckAndSetDefaults method on the Config struct that sets it to the default value (real clock in this case) if it wasn't specified. For example:

```go
type Config {
  Clock clockwork.Clock
}

func (c *Config) CheckAndSetDefaults() error {
  if c.Clock == nil {
    c.Clock = clockwork.NewRealClock()
  }
  return nil
}

func NewTimeline(ctx context.Context, config Config) (xxx, error) {
  if err := config.CheckAndSetDefaults(); err != nil {
    return nil, trace.Wrap(err)
  }
  ...
}

timeline, err := sqlite.NewTimeline(ctx, sqlite.Config{
  DBPath: xxx,
  RetentionDuration: yyy,
})
```

This way the caller doesn't have to pass it explicitly but the tests can still override it.

agent/agent.go
```diff
@@ -210,10 +227,12 @@ type agent struct {
 	// done is a channel used for cleanup.
 	done chan struct{}
 
+	// TODO: do we need separate clocks here?
```
r0mant: That's exactly the question I have reading this 😄maybe they were introduced for tests? I would take a look and get rid of extra clocks if there's no real need for it.

agent/agent.go
```diff
@@ -588,79 +578,113 @@ func (r *agent) getStatusFrom(ctx context.Context, member serf.Member, respc cha
...
+func (r *agent) timelineUpdateLoop() {
+	ticker := r.timelineClock.NewTicker(statusUpdateTimeout)
+	defer ticker.Stop()
 
-// recentStatus returns the last known cluster status.
-func (r *agent) recentStatus() (status *pb.SystemStatus, err error) {
-	status, err = r.cache.RecentStatus()
-	if err == nil && status == nil {
-		status = pb.EmptyStatus()
+	for range ticker.Chan() {
```
r0mant: Similar to what I mentioned in the previous PR, I would rewrite this loop to use select statement and accept the context in this method so it can be canceled by the caller.

agent/agent.go
```diff
@@ -588,79 +578,113 @@ func (r *agent) getStatusFrom(ctx context.Context, member serf.Member, respc cha
...
+		if r.isDone() {
+			return
+		}
+
+		if err := r.updateTimeline(); err != nil {
+			log.WithError(err).Warnf("Error updating timeline.")
```
r0mant: Nit: Just `Warn()`.

agent/agent.go
```diff
@@ -588,79 +578,113 @@ func (r *agent) getStatusFrom(ctx context.Context, member serf.Member, respc cha
...
 
-// recentLocalStatus returns the last known local node status.
-func (r *agent) recentLocalStatus() *pb.NodeStatus {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-	return r.localStatus
-}
+// updateTimeline updates the timeline with any new status changes.
+func (r *agent) updateTimeline() error {
+	ctx, cancel := context.WithTimeout(context.Background(), r.statusQueryReplyTimeout)
```
r0mant: Pass the context to this method. If you address my comment above about using select statement, you'll have the context in the timelineUpdateLoop and be able to pass it here as well.

agent/agent.go
```diff
@@ -588,79 +578,113 @@ func (r *agent) getStatusFrom(ctx context.Context, member serf.Member, respc cha
...
+
+	// Update local status changes
+	if status := r.LocalStatus(); status != nil {
+		if err := r.Timeline.RecordStatus(ctx, status); err != nil {
+			return trace.Wrap(err)
+		}
+	}
+
+	// Collect timeline from another cluster member
+	timeline, err := r.collectTimeline(ctx)
```
r0mant: Didn't we want to have all nodes send their timeline events to the masters so master nodes will have a consistent view of the entire timeline?

agent/agent.go
```diff
@@ -588,79 +578,113 @@ func (r *agent) getStatusFrom(ctx context.Context, member serf.Member, respc cha
...
-	case MemberLeaving:
-		return pb.MemberStatus_Leaving
-	case MemberLeft:
-		return pb.MemberStatus_Left
-	case MemberFailed:
-		return pb.MemberStatus_Failed
-	}
-	return pb.MemberStatus_None
+	// Merge timelines
+	if timeline != nil {
```
r0mant: When can this be nil?

agent/agent.go
```diff
@@ -588,79 +578,113 @@ func (r *agent) getStatusFrom(ctx context.Context, member serf.Member, respc cha
...
+	return nil
 }
 
-// unknownNodeStatus creates an `unknown` node status for a node specified with member.
-func unknownNodeStatus(member *serf.Member) *pb.NodeStatus {
-	return &pb.NodeStatus{
-		Name:         member.Name,
-		Status:       pb.NodeStatus_Unknown,
-		MemberStatus: statusFromMember(member),
+// collectTimeline collects timeline from a random member.
```
r0mant: To be honest, I'm not a big fan of using randomness here. I'm not sure using random will even qualify as an eventually consistent algorithm.

The way I see it, we have two possible approaches to synchronizing the timeline we have to pick between: "pull" vs "push".

The "pull" approach is what you implemented here but instead of using random, I think that at the very least it should use a some sort of round-robin so it is guaranteed that we eventually grab the timeline from all nodes in a predictable fashion. The upside of this approach is that the node that does it (I'm assuming it only runs on masters) has full control over how quickly it iterates and can balance the workload. The downside is that on large clusters it can take a considerable amount of time to collect timeline from all nodes and when processing the last node, the status of the first one may have already changed (I guess this is what constitutes the "eventual" consistency property).

The "push" approach is what I mentioned in the comment above and what I believe we discussed in the design doc. In this scenario each node sends events to all master nodes as soon as they happen (can use some batching). With this approach the timeline on masters will always be consistent but the downside is that all nodes send their stuff to masters. However these events don't happen often so I don't think performance/scalability would be a concern.

We should also factor in various implementation details as well obviously, such as that if a node failed to send events, it should do retries, etc.

I'm personally leaning towards the "push" approach, just because it will make the timeline a much more useful real-time troubleshooting tool, but let me know what you think. cc @a-palchikov @knisbet as well.

bernardjkim: I guess you're right that using random would disqualify this algorithm from being eventually consistent. Will change to a round robin approach.

I agree that the "push" approach will be more efficient. I am working on implementing this feature incrementally, so I decided to go with the "pull" approach first because it was easier to implement. After testing for correctness I was planning on optimizing.

PR #144, opened 2020-01-16
Refactor

bernardjkim:
Refactored some of the agent pkg code before adding additional test cases.
Still some more code that needs to be reorganized, but will come back after finishing up gravity history command.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -85,8 +85,8 @@ type Config struct {
 	// KeyFile specifies the path to TLS certificate key file
 	KeyFile string
 
-	// RPC address of local serf node.
-	SerfRPCAddr string
+	// SerfConfig specifies serf client configuration.
+	SerfConfig serf.Config
 
 	// Address to listen on for web interface and telemetry for Prometheus metrics.
 	MetricsAddr string
@@ -100,11 +100,8 @@ type Config struct {
 	// Tags is a trivial means for adding extra semantic information to an agent.
 	Tags map[string]string
 
-	// DBPath specifies the location of the database to be used to store timeline.
-	DBPath string
-
-	// RetentionDuration specifies the duration to store timeline events.
-	RetentionDuration time.Duration
+	// TimelineConfig specifies sqlite timeline configuration.
+	TimelineConfig sqlite.Config
 
 	// Clock to be used for internal time keeping.
 	Clock clockwork.Clock
@@ -113,125 +110,44 @@ type Config struct {
 	cache.Cache
 }
 
-// New creates an instance of an agent based on configuration options given in config.
-func New(config *Config) (Agent, error) {
-	if err := config.CheckAndSetDefaults(); err != nil {
-		return nil, trace.Wrap(err)
-	}
-
-	clientConfig := serf.Config{
-		Addr: config.SerfRPCAddr,
-	}
-	client, err := NewSerfClient(clientConfig)
-	if err != nil {
-		return nil, trace.Wrap(err, "failed to connect to serf")
-	}
-
-	err = client.UpdateTags(config.Tags, nil)
-	if err != nil {
-		return nil, trace.Wrap(err, "failed to update serf agent tags")
-	}
-
-	metricsListener, err := net.Listen("tcp", config.MetricsAddr)
-	if err != nil {
-		return nil, trace.Wrap(err, "failed to serve prometheus metrics")
-	}
-
-	ctx, cancel := context.WithTimeout(context.Background(), timelineInitTimeout)
-	defer cancel()
-
-	sqliteConfig := sqlite.Config{
-		DBPath:            config.DBPath,
-		RetentionDuration: config.RetentionDuration,
-		Clock:             config.Clock,
-	}
-	timeline, err := sqlite.NewTimeline(ctx, sqliteConfig)
-	if err != nil {
-		return nil, trace.Wrap(err, "failed to initialize timeline")
-	}
-
-	agent := &agent{
-		Config:                  *config,
-		SerfClient:              client,
-		Timeline:                timeline,
-		name:                    config.Name,
-		cache:                   config.Cache,
-		dialRPC:                 DefaultDialRPC(config.CAFile, config.CertFile, config.KeyFile),
-		statusClock:             config.Clock,
-		recycleClock:            config.Clock,
-		timelineClock:           config.Clock,
-		statusQueryReplyTimeout: statusQueryReplyTimeout,
-		localStatus:             emptyNodeStatus(config.Name),
-		metricsListener:         metricsListener,
-		done:                    make(chan struct{}),
-	}
-
-	agent.rpc, err = newRPCServer(agent, config.CAFile, config.CertFile, config.KeyFile, config.RPCAddrs)
-	if err != nil {
-		return nil, trace.Wrap(err, "failed to create RPC server")
-	}
-	return agent, nil
-}
-
 // CheckAndSetDefaults validates this configuration object.
 // Config values that were not specified will be set to their default values if
 // available.
 func (r *Config) CheckAndSetDefaults() error {
 	var errors []error
-
 	if r.CAFile == "" {
 		errors = append(errors, trace.BadParameter("certificate authority file must be provided"))
 	}
-
 	if r.CertFile == "" {
 		errors = append(errors, trace.BadParameter("certificate must be provided"))
 	}
-
 	if r.KeyFile == "" {
 		errors = append(errors, trace.BadParameter("certificate key must be provided"))
 	}
-
 	if r.Name == "" {
 		errors = append(errors, trace.BadParameter("agent name cannot be empty"))
 	}
-
 	if len(r.RPCAddrs) == 0 {
 		errors = append(errors, trace.BadParameter("at least one RPC address must be provided"))
 	}
-
-	if r.DBPath == "" {
-		errors = append(errors, trace.BadParameter("sqlite database path must be provided"))
-	}
-
 	if r.Tags == nil {
 		r.Tags = make(map[string]string)
 	}
-
 	if r.Clock == nil {
 		r.Clock = clockwork.NewRealClock()
 	}
-
 	return trace.NewAggregate(errors...)
 }
 
 type agent struct {
+	sync.Mutex
 	health.Checkers
 
 	metricsListener net.Listener
 
-	// SerfClient provides access to the serf agent.
-	SerfClient SerfClient
-
-	// Name of this agent.  Must be the same as the serf agent's name
-	// running on the same node.
-	name string
-
 	// RPC server used by agent for client communication as well as
 	// status sync with other agents.
 	rpc RPCServer
-
-	// cache persists node status history.
-	cache cache.Cache
 
 	// dialRPC is a factory function to create clients to other agents.
 	// If future, agent address discovery will happen through serf.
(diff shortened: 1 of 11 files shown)


agent/agent.go
```diff
@@ -85,8 +85,8 @@ type Config struct {
 	// KeyFile specifies the path to TLS certificate key file
 	KeyFile string
 
-	// RPC address of local serf node.
-	SerfRPCAddr string
+	// SerfConfig specifies serf client configuration.
```
bernardjkim: Is it ok to nest configs?

agent/agent.go
```diff
@@ -113,143 +110,52 @@ type Config struct {
...
-	// SerfClient provides access to the serf agent.
-	SerfClient SerfClient
-
-	// Name of this agent.  Must be the same as the serf agent's name
-	// running on the same node.
-	name string
-
 	// RPC server used by agent for client communication as well as
 	// status sync with other agents.
 	rpc RPCServer
```
bernardjkim: Does agent need a reference to the RPC server?
From what I can see agent only uses this reference to stop the server in the `Close` method.

agent/agent.go
```diff
@@ -275,25 +248,29 @@ func (r *agent) GetConfig() Config {
...
-			}
-		}()
+// serveMetrics registers the prometheus metrics handler and starts accepting
+// connections on the metrics listener.
+func (r *agent) serveMetrics(ctx context.Context) {
+	if r.metricsListener == nil {
+		return
 	}
 
-	select {
```
bernardjkim: Wasn't exactly sure what the purpose of this code was.
When do we care if `http.Serve` returns an error within the first second?

agent/server.go
```diff
@@ -147,26 +178,37 @@ func newRPCServer(agent *agent, caFile, certFile, keyFile string, rpcAddrs []str
 
 	// handler is a multiplexer for both gRPC and HTTPS queries.
 	// The HTTPS endpoint returns the cluster status as JSON
+	// TODO: why does server need to handle both gRPC and HTTPS queries?
```
bernardjkim: When do we use HTTPS instead of gRPC?

agent/server.go
```diff
@@ -147,26 +178,37 @@ func newRPCServer(agent *agent, caFile, certFile, keyFile string, rpcAddrs []str
...
 	handler := grpcHandlerFunc(server, healthzHandler)
 
 	for _, addr := range rpcAddrs {
-		go func(address string) {
-			if err := serve(address, certFile, keyFile, tlsConfig, handler); err != nil {
-				log.WithError(err).Errorf("Failed to serve on %v.", address)
+		srv := newHTTPServer(addr, tlsConfig, handler)
+		server.httpServers = append(server.httpServers, srv)
```
bernardjkim: http.Servers were being started but never closed/shutdown.

agent/server_test.go
```diff
@@ -289,109 +213,88 @@ func (r *AgentSuite) TestAgentProvidesStatus(c *C) {
...
-		testCase.status.Timestamp = pb.NewTimeToProto(clock.Now())
-		for _, node := range resp.Status.Nodes {
-			sort.Sort(byChecker(node.Probes))
-		}
-		c.Assert(resp.Status, test.DeepCompare, &testCase.status, comment)
-		rpcPort = rpcPort + 1
-	}
-}
-
-func (r *AgentSuite) TestRecyclesCache(c *C) {
```
bernardjkim: Unnecessary test, or should be unit test for cache pkg.

agent/server.go
```diff
@@ -105,6 +108,34 @@ func (r *server) UpdateTimeline(ctx context.Context, req *pb.UpdateRequest) (*pb
 	return &pb.UpdateResponse{}, nil
 }
 
+// Stop stops the grpc server and any additional http servers.
+// TODO: modify Stop to return error
+func (r *server) Stop() {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
a-palchikov: Why not take the context as a parameter?

agent/server.go
```diff
@@ -105,6 +108,34 @@ func (r *server) UpdateTimeline(ctx context.Context, req *pb.UpdateRequest) (*pb
...
+	defer cancel()
+
+	if err := r.stopHTTPServers(ctx); err != nil {
+		log.WithError(err).Error("All HTTP servers failed to shutdown.")
```
a-palchikov: Technically not true about `all` as the method might return errors for a subset.

agent/server.go
a-palchikov: We use HTTPS to support a simple curl-style query for status.

agent/server.go
```diff
@@ -52,7 +54,8 @@ type RPCServer interface {
 // server implements RPCServer for an agent.
 type server struct {
 	*grpc.Server
-	agent *agent
+	agent       *agent
+	httpServers []*http.Server
```
a-palchikov: It looks like you're only saving enough to be able to call `Shutdown` and don't need anything beyond that - factor these into a slice of `gracefulCloser`?
```go
type server struct {
	// ...
	closers []gracefulCloser
}

type gracefulCloser interface {
	Shutdown(context.Context)
}
```

agent/server.go
```diff
@@ -147,26 +178,37 @@ func newRPCServer(agent *agent, caFile, certFile, keyFile string, rpcAddrs []str
...
+
+		// TODO: separate Start function to start listening.
+		go func(srv *http.Server) {
+			err := srv.ListenAndServeTLS(certFile, keyFile)
+			if err == http.ErrServerClosed {
+				log.WithError(err).Debug("Server has been shutdown/closed.")
+				return
+			}
+			if err != nil {
+				log.WithError(err).Errorf("Failed to serve on %v.", srv.Addr)
```
a-palchikov: this block would ideally trigger agent process to fail as well.

agent/status_test.go
```diff
@@ -0,0 +1,136 @@
...
+					Status:   pb.Probe_Failed,
+					Severity: pb.Probe_Critical,
+					Error:    "not available",
+				},
+			},
+		},
+	}
+
+	setSystemStatus(resp.Status, []serf.Member{{Name: "foo"}, {Name: "bar"}})
+	c.Assert(resp.Status.Status, Equals, pb.SystemStatus_Degraded)
```
a-palchikov: This should check the full status and not a part of it - or is supposed to exercise just a subset of `setSystemStatus` functionality? The full test is always preferred over the partial one as it will exclude any unexpected changes.
ditto for tests below

agent/server.go
bernardjkim: server.Addr is being used when logging errors which might be useful.
Will keep it as is for now, but will look to clean it up later.

agent/server.go
bernardjkim: I was thinking it would be better to move this block into a separate `start` function.
We can then implement some sort of retry strategy and either wait and return any errors or return an error channel for the caller to wait to handle server shutdown.

agent/server.go
bernardjkim: Added TODO. Will add in another round of refactoring.

lib/history/memory/inmemory.go
```diff
@@ -102,3 +112,16 @@ func (t *Timeline) getFilteredEvents(params map[string]string) (events []*pb.Tim
 	// TODO: for now just return the unfiltered events.
 	return t.events
 }
+
+// filterDuplicates removes duplicate events.
+func (t *Timeline) filterDuplicates() {
+	set := make(map[*pb.TimelineEvent]struct{})
```
r0mant: Hm, so what is a "duplicate" event in this case? Since this holds a pointer to a TimelineEvent, it means that a duplicate is exact same event pointer? When/how can this happen?

lib/history/memory/inmemory.go
```diff
@@ -102,3 +112,16 @@ func (t *Timeline) getFilteredEvents(params map[string]string) (events []*pb.Tim
...
+	filteredEvents := make([]*pb.TimelineEvent, 0, t.capacity)
+	for _, event := range t.events {
+		if _, ok := set[event]; !ok {
```
r0mant: If the inmemory timeline would hold a map of events instead of a slice, you wouldn't need to filter anything in the first place.

agent/server.go
```diff
@@ -105,6 +108,35 @@ func (r *server) UpdateTimeline(ctx context.Context, req *pb.UpdateRequest) (*pb
...
+
+	r.Server.Stop()
+}
+
+// stopHTTPServers shuts down all listening http servers.
+func (r *server) stopHTTPServers(ctx context.Context) error {
+	var errors []error
+	for _, srv := range r.httpServers {
+		err := srv.Shutdown(ctx)
+		if utils.IsContextDone(ctx) {
```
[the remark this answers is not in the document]
bernardjkim: I was thinking that if the context is done every other call to Shutdown will return immediately and is wasting resources.

I guess we don't need to handle context is done case separately so that we can log each server that failed to shutdown before context is done.

lib/history/memory/inmemory.go
bernardjkim: Yeah, so this was a bug.
In order to compare TimelineEvents they first need to be converted into a comparable struct.

lib/history/memory/inmemory.go
bernardjkim: Looks like it might be worth to leave it as a slice. We are removing the oldest events if the timeline reaches max capacity. Doing this with a slice is fairly simple. 

If we were storing events in a map, we will need to iterate over the events to find the oldest events.
We can optimize by using a combination of a linked list and a map, but that seems unnecessary right now since this implementation is really only meant to be used for dev/testing.

agent/agent.go
a-palchikov: I think the intention was to handle the case of http failing to `Serve` immediately which should have failed the whole method - now, this case is simply ignored.

agent/agent.go
a-palchikov: You mean embed config into one another? Perfect use of aggregation imo.

agent/agent.go
a-palchikov: Not sure what you mean?
If you meant it could do away with a simple stopper interface - then sure, it could.

bernardjkim: Sorry, should have elaborate a bit more on this.

- agent stores a pointer to the RPC server.
- RPC server also stores a pointer to the agent.
- ^ Unnecessary circular reference.
- The server de-references the agent when handling RPCs, so the server needs a pointer to the agent.
- The agent only holds a pointer to the server so that it can shutdown the server.

```golang
func runAgent(config *agent.Config, monitoringConfig *config, peers []string) error {
	...
	monitoringAgent, err := agent.New(config)
	if err != nil {
		return trace.Wrap(err)
	}
	defer monitoringAgent.Close()


	if err = addCheckers(monitoringAgent, monitoringConfig); err != nil {
		return trace.Wrap(err)
	}
	if err = monitoringAgent.Start(); err != nil {
		return trace.Wrap(err)
	}
	if len(peers) > 0 {
		if err = monitoringAgent.Join(peers); err != nil {
			return trace.Wrap(err, "failed to join serf cluster")
		}
	}
        ...
}
```
Currently it looks like the main function handles agent management (init/start/stop).
But I feel like it would make more sense to have to have it the other way. Main should handle server management (init/start/stop) and have the server manage agent internally. This way, agent does not need to store a pointer to the server.

```golang
func runServer(...) error {
    monitoringAgent, err := agent.New(...)
    if err != nil {
        return trace.Wrap(err)
    }
    // Additional agent init
    ...

    server, err := newRPCServer(agent, ...)
    if err != nil {
        return trace.Wrap(err)
    }
    // Handle server start/stop
    ....
}
```

PR #148, opened 2020-01-28
[6.1.x] SELinux support

a-palchikov:


Gopkg.toml
```diff
@@ -34,7 +34,8 @@
 
 [[override]]
   name = "github.com/mitchellh/go-ps"
-  branch = "master"
+  source = "github.com/gravitational/go-ps"
```
r0mant: Do you think it's worth trying submitting the PR upstream? We'll soon fork all our dependencies :)

monitoring/port.go
```diff
@@ -108,10 +107,8 @@ func (c *portChecker) checkProcess(proc process, reporter health.Reporter) bool
 			continue
 		}
 		switch proc.socket.state() {
-		case TimeWait:
-			// ignore sockets in time-wait state since they're going
-			// away soon
-			log.Debugf("Ignoring %v for program %q(pid=%v).", formatSocket(proc.socket), proc.name, proc.pid)
```
r0mant: I would keep this log entry.

monitoring/procfs_net.go
```diff
@@ -63,8 +63,11 @@ func (r portCollector) sockets(fetchSockets ...socketGetterFunc) (ret []process,
 
 	for _, socket := range sockets {
 		proc, err := r.findProcessByInode(socket.inode())
-		if err != nil {
-			log.Warn(err.Error())
+		if err != nil && socket.inode() != "0" && !trace.IsNotFound(err) {
```
r0mant: Could you either add a comment explaining what this condition checks for, or move it into a helper method with descriptive name?

Gopkg.toml
a-palchikov: https://github.com/mitchellh/go-ps/pull/39

PR #149, opened 2020-01-29
Timeline optimizations

bernardjkim:
Needed to restructure the code a little bit before implementing changes:
- Client is no longer a part of the `agent` pkg. It has been moved to a separate `client` pkg.
- Timeline Interface has been modified so that `GetEvents` returns a **sorted** list of events by timestamp.

RPC:
- Add `LastSeen` rpc that can be used to query the last seen timestamp for a specified node.

Agent changes:
- Master nodes now keep track of last seen timestamps for each member of the cluster. This information will be used to filter out events that have already been recorded on the master.
- Only master nodes keep track of the overall timeline.
- All agents now keep track of a local timeline. New events will be pushed to all master nodes in the cluster.


agent/agent.go
```diff
@@ -103,6 +105,9 @@ type Config struct {
 	// TimelineConfig specifies sqlite timeline configuration.
 	TimelineConfig sqlite.Config
 
+	// LocalTimelineConfig specifies local timeline configuration.
+	LocalTimelineConfig sqlite.Config
```
r0mant: How's this different from `TimelineConfig` right above?

agent/agent.go
```diff
@@ -164,20 +169,23 @@ type agent struct {
...
 
-	// memberIndex is used to select the next member to collect timeline data
-	// from.
-	// Type uint16 so limited to 0 - 65535. Assuming clusters will not have
-	// more than that many members.
-	memberIndex uint16
+	// lastSeen keeps track of the last seen timestamp from a cluster member.
+	// The last seen timestamp can be queried by a member and be used to
+	// filter out events that have already been recorded by this member.
+	lastSeen map[string]time.Time
```
r0mant: Do members ever get removed from this map? Wonder if you considered using TTL map here.

agent/agent.go
```diff
@@ -164,20 +169,23 @@ type agent struct {
...
 	// SerfClient provides access to the serf agent.
 	SerfClient SerfClient
 
-	// Timeline keeps track of status change events.
+	// Timeline keeps track of all timeline events in the cluster. This timeline
+	// is only used on members that have the role 'master'.
 	Timeline history.Timeline
+
+	// LocalTimeline keeps track of local timeline events.
+	LocalTimeline history.Timeline
```
r0mant: I'm curious why agent needs to keep 2 timelines.

agent/agent.go
```diff
@@ -528,153 +571,117 @@ L:
...
-// the local node.
-func (r *agent) updateLocalChanges(ctx context.Context) error {
-	ctxUpdate, cancel := context.WithTimeout(ctx, updateTimelineTimeout)
-	defer cancel()
-
-	if err := r.Timeline.RecordStatus(ctxUpdate, r.LocalStatus()); err != nil {
-		return trace.Wrap(err)
+	// TODO: async
+	for _, member := range members {
+		if role, ok := member.Tags["role"]; ok && Role(role) == RoleMaster {
```
r0mant: I would either consider moving this condition into a helper method, or even adding a convenient helper `Masters()` method on the serf client.

Also, do we need to only consider "active" members here?

agent/agent.go
```diff
@@ -528,153 +571,117 @@ L:
...
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	defer client.Close()
 
-	ctxUpdate, cancel := context.WithTimeout(ctx, updateTimelineTimeout)
-	defer cancel()
-	if err := r.Timeline.RecordTimeline(ctxUpdate, timeline); err != nil {
+	// Filter events by last seen timestamp. Only push unrecorded events.
+	resp, err := client.LastSeen(ctx, &pb.LastSeenRequest{Name: r.Name})
```
r0mant: So do I understand correctly that "last seen" timestamp is stored on master nodes and indicates the last time they received an event from a particular node?

agent/server_test.go
```diff
@@ -567,23 +594,66 @@ func (r *AgentSuite) newAgent(config testAgentConfig) *agent {
...
+	}
+
+	creds := credentials.NewTLS(&tls.Config{
+		RootCAs:      certPool,
+		Certificates: []tls.Certificate{cert},
+		MinVersion:   tls.VersionTLS12,
+		ServerName:   serverName,
+		// Use TLS Modern capability suites
+		// https://wiki.mozilla.org/Security/Server_Side_TLS
+		CipherSuites: []uint16{
```
r0mant: Nit: I'd move this list to constants/defaults.

lib/client/client.go
```diff
@@ -49,40 +90,35 @@ type client struct {
...
-		return nil, trace.Wrap(err, "failed to append certificates from %v", caFile)
+		return nil, trace.Wrap(err, "failed to append certificates from %v", config.CAFile)
 	}
 
 	creds := credentials.NewTLS(&tls.Config{
 		RootCAs:      certPool,
 		Certificates: []tls.Certificate{cert},
 		MinVersion:   tls.VersionTLS12,
-		ServerName:   serverName,
+		ServerName:   "",
```
r0mant: Why this change?

agent/agent.go
bernardjkim: The local timeline config just has a different sqlite db path. I guess we can just have a single config for both that specifies a directory path and have the overall cluster timeline and local timeline be constant file names in that path.

agent/agent.go
bernardjkim: The 'last seen' is the timestamp of the latest recorded event for a particular node. Will update doc to clarify.

lib/client/client.go
bernardjkim: The option to override `serverName` was only available for unit tests. In normal use `NewClient` was initializing client with no `ServerName`.
Moved this functionality out of `client.go` and into the test file 'client_test.go` to simplify the code.

agent/agent.go
bernardjkim: Two timelines are needed since only the master nodes are expected to store the overall cluster timeline.
The local timeline is used internally, and is only used to cache local events before pushing to master nodes.
Only the master nodes need to keep an additional cluster timeline, so we can add some conditionals to remove the cluster timeline from non master nodes.

PR #150, opened 2020-02-03
More testing

bernardjkim:
Added a `membership` package: 
- Provides `ClusterMembership` interface, renamed from `SerfClient`. Also provides `ClusterMember` interface.
- `SerfMember` struct implements `ClusterMember` and embeds `serf.Member`.

Agent changes:
- Agent now interfaces with `ClusterMembership` and `ClusterMember`. This makes it easier to mock functionality when testing.
- Renamed `server_test.go` to `agent_test.go`.
- No longer need to start up rpc server when testing.

Sqlite testing:
- Verify sqlite database can be created in a new directory.
- Verify timeline can be reinitialized with a preexisting database.


agent/agent.go
```diff
@@ -652,8 +654,11 @@ func (r *agent) collectLocalStatus(ctx context.Context, local *serf.Member) (sta
 }
 
 // getLocalStatus obtains local node status.
-func (r *agent) getLocalStatus(ctx context.Context, local serf.Member, respc chan<- *statusResponse) {
-	status, err := r.collectLocalStatus(ctx, &local)
+func (r *agent) getLocalStatus(ctx context.Context, respc chan<- *statusResponse) {
+	// TODO: restructure code so that local member is not needed here.
+	local, _ := r.ClusterMembership.FindMember(r.Name)
```
r0mant: Do not ignore the error.

lib/membership/membership.go
```diff
@@ -0,0 +1,60 @@
...
+	// Close closes the client.
+	Close() error
+	// Join attempts to join an existing cluster identified by peers.
+	// Replay controls if previous user events are replayed once this node has joined the cluster.
+	// Returns the number of nodes joined.
+	Join(peers []string, replay bool) (int, error)
+	// UpdateTags will modify the tags on a running member.
+	UpdateTags(tags map[string]string, delTags []string) error
+	// GetCoordinate returns the Serf Coordinate for a specific node
+	// GetCoordinate(node string) (*coordinate.Coordinate, error)
```
r0mant: Why is this commented out?

lib/membership/constants.go
```diff
@@ -0,0 +1,28 @@
...
+*/
+
+package membership
+
+// MemberStatus describes the state of a serf node.
+type MemberStatus string
+
+// Possible membership statuses.
+const (
+	MemberAlive   MemberStatus = "alive"
```
r0mant: Nit: Could you add godocs on these fields?

lib/membership/membership.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// ClusterMember is interface to interact with a cluster member.
+type ClusterMember interface {
+	// Dial attempts to create client connect to member.
+	Dial(ctx context.Context, caFile, certFile, keyFile string) (client.Client, error)
```
r0mant: Maybe it's just me but I find it a bit weird that "Dial" is a part of the cluster member interface. This means that when the caller deals with it, ClusterMember instance can be in the "non initialized" state i.e. require to call Dial first. Does this have to be a part of the interface? I would consider moving this into a separate public method that returns a ClusterMember instance.

lib/membership/serf.go
```diff
@@ -0,0 +1,222 @@
...
+
+// SerfMember embeds serf.Member and implements ClusterMember.
+type SerfMember struct {
+	*serf.Member
+}
+
+// Dial attempts to create client connection to the serf member.
+func (r SerfMember) Dial(ctx context.Context, caFile, certFile, keyFile string) (client.Client, error) {
+	// RPCPort defines default rpc port
+	const RPCPort = 7575
```
r0mant: Don't we have a constant for this elsewhere?

lib/membership/membership.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Package membership provides interface for cluster membership management.
+package membership
+
+import (
+	"net"
+
+	"github.com/gravitational/satellite/lib/client"
+
+	"golang.org/x/net/context"
```
r0mant: Why not builtin `context` package?

lib/membership/membership.go
bernardjkim: > This means that when the caller deals with it, ClusterMember instance can be in the "non initialized" state i.e. require to call Dial first.

Can you clarify what problems can occur when dealing with a non initialized member?

Would it make sense to have a `Dialer/RPCDialer` interface but still have `SerfMember` implement `Dialer`?

agent/agent.go
```diff
@@ -183,13 +160,13 @@ type agent struct {
...
-	lastSeen ttlmap.TTLMap
+	lastSeen *ttlmap.TTLMap
 
 	// Config is the agent configuration.
 	Config
 
-	// SerfClient provides access to the serf agent.
-	SerfClient SerfClient
+	// ClusterMembership provides access to cluster membership service.
+	ClusterMembership membership.ClusterMembership
```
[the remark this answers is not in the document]
bernardjkim: - I think it makes sense to have the interfaces be defined locally in the `agent` pkg and leave only the necessary functions. Will make necessary changes.
- I think the code structure definitely has room to improve. However, I think this approach simplifies the testability of agent which is more important right now. Will work on code structure in future refactoring.

PR #158, opened 2020-02-12
Updated & revendored to Kubernetes 1.17.3

eldios:
Updated & revendored to Kubernetes 1.17.3


eldios:
re: https://github.com/gravitational/gravity/issues/1102

eldios:
I was thinking about possible API issues in future with deprecation getting in the way.. :thinking: 

I guess we can keep it back for now until we find such issues?

eldios:
won't update as per @r0mant suggestion.

PR #164, opened 2020-02-25
Add nethealth checker

bernardjkim:
This PR adds the nethealth checker. [Design Doc](https://docs.google.com/document/d/1zJGGYLPWmNioanIpzuvtjZNffgFzJ-Ezkv3vP3kLKLM/edit#heading=h.208t1zijroe)

### Rationale
Satellite does not currently have a health check in place to verify network communication between peers. Blocked UDP traffic due to a firewall or other network issues may go undetected.

### Implementation
The [nethealth](https://github.com/gravitational/satellite/pull/113) service exposes a counter for the number of echo requests and timeouts from peers in a cluster. This checker pulls metrics from the nethealth service and verifies that the network communication between peers is functional. The timeout stats are recorded as a short time series data interval containing a user specified number of data points. If the packet loss percentage is above a specified threshold at each data point, the network will be considered unhealthy and will result in a failed check.

### DIFF
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -40,9 +40,11 @@ type NethealthConfig struct {
 	// Should match the k8s node name.
 	NodeName string
 
-	// NethealthServiceAddr specifies the nethealth service address used to
-	// to collect network metrics.
-	NethealthServiceAddr string
+	// NethealthIP specifies the IP address of the nethealth pod.
+	NethealthIP string
+
+	// NethealthPort specifies the port that nethealth is listening on.
+	NethealthPort int
 
 	// SeriesCapacity specifies the max number of data points to store in a time
 	// series interval.
@@ -56,8 +58,14 @@ func (c *NethealthConfig) CheckAndSetDefaults() error {
 	if c.NodeName == "" {
 		errors = append(errors, trace.BadParameter("node name must be provided"))
 	}
-	if c.NethealthServiceAddr == "" {
-		errors = append(errors, trace.BadParameter("nethealth service address must be provided"))
+	if c.NethealthIP == "" {
+		errors = append(errors, trace.BadParameter("nethealth ip address must be provided"))
+	}
+	if c.NethealthPort == 0 {
+		c.NethealthPort = defaultNethealthPort
+	}
+	if c.SeriesCapacity < 0 {
+		errors = append(errors, trace.BadParameter("timeout series capacity cannot be < 0"))
 	}
 	if c.SeriesCapacity == 0 {
 		c.SeriesCapacity = defaultSeriesCapacity
@@ -81,6 +89,10 @@ type nethealthChecker struct {
 
 // NewNethealthChecker returns a new nethealth checker.
 func NewNethealthChecker(config NethealthConfig) (*nethealthChecker, error) {
+	if err := config.CheckAndSetDefaults(); err != nil {
+		return nil, trace.Wrap(err)
+	}
+
 	return &nethealthChecker{
 		timeoutStats:    holster.NewTTLMap(timeoutStatsCapacity),
 		NethealthConfig: config,
@@ -104,7 +116,7 @@ func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter) 
 }
 
 func (c *nethealthChecker) check(ctx context.Context) error {
-	b, err := fetchMetrics(ctx, c.NethealthServiceAddr)
+	b, err := fetchMetrics(ctx, c.NethealthIP, c.NethealthPort)
 	if err != nil {
 		return trace.Wrap(err, "failed to fetch metrics")
 	}
@@ -131,14 +143,9 @@ func (c *nethealthChecker) check(ctx context.Context) error {
 func (c *nethealthChecker) updateStats(metrics []*dto.Metric) (updated []string, err error) {
 	var errors []error
 	for _, metric := range metrics {
-		nodeName, peerName, err := getNames(metric.GetLabel())
+		peerName, err := getPeerName(metric.GetLabel())
 		if err != nil {
 			errors = append(errors, trace.Wrap(err))
-			continue
-		}
-
-		// Only update metrics relevant to this node.
-		if c.NodeName != nodeName {
 			continue
 		}
 
@@ -149,7 +156,7 @@ func (c *nethealthChecker) updateStats(metrics []*dto.Metric) (updated []string,
 		}
 
 		// Keep only the last `seriesCapacity` number of data points.
-		if len(series) >= c.SeriesCapacity {
+		if len(series) >= c.SeriesCapacity && len(series) > 0 {
 			series = series[1:]
 		}
 		series = append(series, int64(metric.GetCounter().GetValue()))
@@ -158,7 +165,7 @@ func (c *nethealthChecker) updateStats(metrics []*dto.Metric) (updated []string,
 			errors = append(errors, trace.Wrap(err))
 		}
 
-		// Record updated nodes to be returned for use in later network verification step.
+		// Record updated nodes to be returned for use in later nethealth verification step.
 		updated = append(updated, peerName)
 	}
 	return updated, trace.NewAggregate(errors...)
@@ -221,8 +228,8 @@ func (c *nethealthChecker) setTimeoutSeries(name string, series []int64) error {
 
 // fetchMetrics collects the network metrics from the nethealth service.
 // Metrics are returned as an array of bytes.
-func fetchMetrics(ctx context.Context, addr string) ([]byte, error) {
-	addr = "http://" + addr
+func fetchMetrics(ctx context.Context, ip string, port int) ([]byte, error) {
+	addr := fmt.Sprintf("http://%s:%d", ip, port)
 
 	client, err := roundtrip.NewClient(addr, "/metrics", roundtrip.HTTPClient(&http.Client{}))
 	if err != nil {
@@ -255,32 +262,26 @@ func parseMetrics(input io.Reader) ([]*dto.Metric, error) {
 	return nil, trace.NotFound("%s metrics not found", requestTimeoutName)
 }
 
-// getNames extracts the 'node_name' and 'peer_name' values from the provided
-// labels.
-func getNames(labels []*dto.LabelPair) (node, peer string, err error) {
+// getPeerName extracts the 'peer_name' value from the provided labels.
+func getPeerName(labels []*dto.LabelPair) (peer string, err error) {
 	for _, label := range labels {
-		name := label.GetName()
-		if peerLabel == name {
-			peer = label.GetValue()
-		}
-		if nodeLabel == name {
-			node = label.GetValue()
-		}
-		if node != "" && peer != "" {
-			return node, peer, nil
+		if peerLabel == label.GetName() {
+			return label.GetValue(), nil
 		}
 	}
-	return "", "", trace.NotFound("unable to find required labels")
+	return "", trace.NotFound("unable to find required peer label")
 }
 
 const (
 	nethealthCheckerID = "nethealth-checker"
 	peerLabel          = "peer_name"
-	nodeLabel          = "node_name"
 
-	// defaultSeriesCapacity  defines the default capacity of a time series
+	// defaultSeriesCapacity defines the default capacity of a time series
 	// interval.
 	defaultSeriesCapacity = 10
+
+	// defaultNethealthPort defines the default nethealth port.
+	defaultNethealthPort = 9801
 
 	// timeoutStatsCapacity sets the number of TTLMaps that can be stored.
 	// This will be the size of the cluster -1.
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -125,7 +125,7 @@ func (c *nethealthChecker) check(ctx context.Context) error {
 }
 
 // getMetrics returns the network metrics from the local nethealth pod.
-func (c *nethealthChecker) getMetrics(ctx context.Context) ([]*dto.Metric, error) {
+func (c *nethealthChecker) getMetrics(ctx context.Context) ([]nethealthMetric, error) {
 	addr, err := c.getNethealthAddr()
 	if err != nil {
 		return nil, trace.Wrap(err, "failed to get local nethealth address")
@@ -147,16 +147,10 @@ func (c *nethealthChecker) getMetrics(ctx context.Context) ([]*dto.Metric, error
 
 // updateStats updates the timeout data with new data points collected in
 // the provided metrics. Returns a list containing the updated keys.
-func (c *nethealthChecker) updateStats(metrics []*dto.Metric) (updated []string, err error) {
+func (c *nethealthChecker) updateStats(metrics []nethealthMetric) (updated []string, err error) {
 	var errors []error
 	for _, metric := range metrics {
-		peerName, err := getPeerName(metric.GetLabel())
-		if err != nil {
-			errors = append(errors, trace.Wrap(err))
-			continue
-		}
-
-		series, err := c.getTimeoutSeries(peerName)
+		series, err := c.getTimeoutSeries(metric.peerName)
 		if err != nil {
 			errors = append(errors, trace.Wrap(err))
 			continue
@@ -166,14 +160,14 @@ func (c *nethealthChecker) updateStats(metrics []*dto.Metric) (updated []string,
 		if len(series) >= c.SeriesCapacity {
 			series = series[1:]
 		}
-		series = append(series, int64(metric.GetCounter().GetValue()))
+		series = append(series, metric.totalTimeout)
 
-		if err := c.setTimeoutSeries(peerName, series); err != nil {
+		if err := c.setTimeoutSeries(metric.peerName, series); err != nil {
 			errors = append(errors, trace.Wrap(err))
 		}
 
 		// Record updated nodes to be returned for use in later nethealth verification step.
-		updated = append(updated, peerName)
+		updated = append(updated, metric.peerName)
 	}
 	return updated, trace.NewAggregate(errors...)
 }
@@ -270,7 +264,7 @@ func fetchMetrics(ctx context.Context, addr string) ([]byte, error) {
 
 // parseMetrics parses input from the provided reader and returns the metrics
 // for the 'nethealth_echo_timeout_total' counter.
-func parseMetrics(input io.Reader) ([]*dto.Metric, error) {
+func parseMetrics(input io.Reader) ([]nethealthMetric, error) {
 	// requestTimeoutName defines the metric family name of the relevant nethealth counter
 	const requestTimeoutName = "nethealth_echo_timeout_total"
 
@@ -280,10 +274,24 @@ func parseMetrics(input io.Reader) ([]*dto.Metric, error) {
 		return nil, trace.Wrap(err)
 	}
 
-	if mf, ok := metricsFamilies[requestTimeoutName]; ok {
-		return mf.GetMetric(), nil
+	mf, ok := metricsFamilies[requestTimeoutName]
+	if !ok {
+		return nil, trace.NotFound("%s metrics not found", requestTimeoutName)
 	}
-	return nil, trace.NotFound("%s metrics not found", requestTimeoutName)
+
+	metrics := make([]nethealthMetric, 0, len(mf.GetMetric()))
+	for _, m := range mf.GetMetric() {
+		peerName, err := getPeerName(m.GetLabel())
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+
+		metrics = append(metrics, nethealthMetric{
+			peerName:     peerName,
+			totalTimeout: int64(m.GetCounter().GetValue()),
+		})
+	}
+	return metrics, nil
 }
 
 // getPeerName extracts the 'peer_name' value from the provided labels.
@@ -294,6 +302,13 @@ func getPeerName(labels []*dto.LabelPair) (peer string, err error) {
 		}
 	}
 	return "", trace.NotFound("unable to find required peer label")
+}
+
+type nethealthMetric struct {
+	// peerName is the node name of the peer.
+	peerName string
+	// totalTimeout is the running total of requests to the peer that have timed out.
+	totalTimeout int64
 }
 
 const (
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -77,9 +77,8 @@ type nethealthChecker struct {
 	seriesCapacity int
 	// lock access to timeoutStats
 	sync.Mutex
-	// netStats maps a peer to a series data interval containing the packet loss
-	// percentage per peer for the set interval.
-	netStats *holster.TTLMap
+	// peerStats maps a peer to its recorded nethealth stats.
+	peerStats netStats
 }
 
 // NewNethealthChecker returns a new nethealth checker.
@@ -93,7 +92,7 @@ func NewNethealthChecker(config NethealthConfig) (*nethealthChecker, error) {
 	seriesCapacity := math.Ceil(float64(config.NetStatsInterval) / float64(agent.StatusUpdateTimeout))
 
 	return &nethealthChecker{
-		netStats:        holster.NewTTLMap(netStatsCapacity),
+		peerStats:       newNetStats(netStatsCapacity),
 		seriesCapacity:  int(seriesCapacity),
 		NethealthConfig: config,
 	}, nil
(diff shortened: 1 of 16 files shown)


knisbet:
So there is an issue with this approach that doesn't appear to have been caught on the design doc, when considering nethealth. By connecting to nethealth.monitoring.svc.cluster.local:9801/metrics, this uses the overlay network that may be blocked, and also accesses a random pod which is a single nodes perspective.

So each satellite agent should monitor the nethealth pod running on the same node, which provides only it's own perspective on connectivity to other nodes. That agent can then relay it's perspective over the peer-to-peer sync (presumably each node is running this check). So the way to do that, would be to instead do a DNS query for any.nethealth.monitoring.svc.cluster.local for the list of every POD providing the service, and then filtering to just the IP that belongs to the same node, and then scraping the metrics from the same node. 

Alternatively, the checker would need a kubernetes client, and to either watch or periodically query the pod objects, as the pod IP addresses can change if the pod is re-created, and then find the pod object with matching labels, that belongs to the same node as the checker. Going the DNS route is lighter weight overall.

r0mant:
@knisbet Wouldn't prometheus' service monitor exhibit the same behavior then? IIRC it also queries a service.

@bernardjkim I would go with the DNS approach Kevin mentioned above, but I think it makes sense  to query the local resolver (127.0.0.2) to ensure that traffic stays local. This means satellite will also need to find and cache information about the pod subnet of the node it's running on upon startup.

knisbet:
The subnet assigned to the host is available from within planet: /run/flannel/subnet.env
flannel writes this configuration when assigning a subnet. In theory it can change, but in practice it shouldn't.

@r0mant Prometheus uses the service to identify pods, but then I believe queries them individually. So you still get per pod/node results from prometheus, but configuration wise it's pointed at a service. Prometheus does exhibit similar limitations, the difference being that prometheus should report scraping failures to pods it's not able to contact, but if the overlay network is totally broken it may provide non intuitive results. The second component of which with the nethealth originally, was that on suspecting an overlay networking problem, the pod logs for nethealth could be observed to provide further insight. For partial interruptions, prometheus + alertmanager are able to report packet loss or totally interruption between any pair of nodes.

bernardjkim:
So right now I'm planning on having planet be responsible for getting the IP address for the local nethealth service. The satellite nethealth checker will just query the metrics from the IP and port provided by planet.

We could do something like:
```golang
func getSubnet() (*net.IPNet, error) {
 	iface, err := net.InterfaceByName("cni0")
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
 	addrs, err := iface.Addrs()
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
	if subnet, ok := addrs[0].(*net.IPNet); ok {
		return subnet, nil
	}
	return nil, trace.BadParameter("failed to get %s subnet", "cni0")
}
```

or:
```golang
func getSubnet() (*net.IPNet, error) {
	const envSubnetPath = "/run/flannel/subnet.env"
	const envSubnetKey = "FLANNEL_SUBNET"

	err := godotenv.Load(envSubnetPath)
	if err != nil {
		return nil, trace.NotFound("unable to find file %s", envSubnetPath)
	}

	subnetString := os.Getenv(envSubnetKey)
	if subnetString == "" {
		return nil, trace.NotFound("%s is not set", envSubnetKey)
	}

	_, subnet, err := net.ParseCIDR(subnetString)
	if err != nil {
		return nil, trace.Wrap(err)
	}

	return subnet, nil
}
```

Then check to see the local subnet contains any of the nethealth service ip addresses:
```golang
func getNethealthIP(host string) (net.IP, error) {
	subnet, err := getSubnet()
	if err != nil {
		return nil, trace.Wrap(err)
	}

	// Lookup all nethealth ip addresses.
	ips, err := net.LookupIP(host) // host = "any.nethealth.monitoring.svc.cluster.local"
	if err != nil {
		return nil, trace.Wrap(err)
	}

	// Return local nethealth ip.
	for _, ip := range ips {
		if subnet.Contains(ip) {
			return ip, nil
		}
	}
	return nil, trace.NotFound("unable to lookup local nethealth ip")
}
```

@knisbet, @r0mant what do you guys think?

r0mant:
The approach with using flannel's configuration to determine the subnet won't work if the flannel is disabled which happens if a user chooses to deploy their own cni plugin. I wonder if there's a way to do this without relying on flannel/interface. I'll think about it.

r0mant:
I mean, one idea which is similar somewhat to what Kevin mentioned above is to use Kubernetes API but instead of watching pods or filtering by labels, the agent can just query the correct pod by the advertise IP of the node (or query all and filter by advertise IP). So:

* Planet agents knows advertise address of the node its running on (or can figure out from /etc/container-environment).
* Then the agent can query the Kubernetes API for pods and find the one which runs on this host (should have .status.hostIP property same as the advertise IP).
* Then the agent makes query to this pod's IP.

One thing to keep in mind here is that on regular nodes "kubectl" does not have permissions to read pods so the agent will need to use k8s client with "kubelet" config (I think there should already be one).

knisbet:
Yea, kubernetes client might be the way to go here if we want to support any potential CNI plugin. We don't necessarily have to though, since we only officially offer support for two plugins (flannel and wormhole).

bernardjkim:
Alright, made some changes in 1488378.
The nethealth checker now queries the kubernetes API for pods and filters using advertised IP.

monitoring/nethealth.go
```diff
@@ -226,11 +235,28 @@ func (c *nethealthChecker) setTimeoutSeries(name string, series []int64) error {
...
+// getNethealthAddr returns the address of the local nethealth pod.
+func (c *nethealthChecker) getNethealthAddr() (string, error) {
+	pods, err := c.Client.CoreV1().Pods(monitoringNamespace).List(metav1.ListOptions{})
+	if err != nil {
+		return "", trace.Wrap(err)
+	}
 
+	// Find nethealth pod with matching host ip address.
+	for _, pod := range pods.Items {
+		if !strings.Contains(pod.GetName(), nethealthName) {
```
a-palchikov: You should list pods using specific label/field selectors to avoid this - otherwise you're running the risk of picking something arbitrary.

monitoring/nethealth.go
```diff
@@ -0,0 +1,293 @@
...
+// nethealthChecker checks network communication between peers.
+type nethealthChecker struct {
+	// lock access to timeoutStats
+	sync.Mutex
+
+	// timeoutStats maps a peer to a time series data interval containing the
+	// number of echo timeouts received by the specific peer for a set interval.
+	timeoutStats *holster.TTLMap
+
+	// NethealthConfig contains caller specified nethealth checker configuration
```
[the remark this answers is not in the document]
bernardjkim: f8d89e9

monitoring/nethealth.go
bernardjkim: Addressed in f8d89e9.
Does this is implementation look correct?

monitoring/nethealth.go
```diff
@@ -0,0 +1,293 @@
...
+		// Record updated nodes to be returned for use in later network verification step.
+		updated = append(updated, peerName)
+	}
+	return updated, trace.NewAggregate(errors...)
+}
+
+// verifyNethealth verifies that the network communication is healthy for the
+// nodes specified by the provided list of names.
+func (c *nethealthChecker) verifyNethealth(names []string) error {
+	var errors []error
```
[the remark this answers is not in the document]
bernardjkim: Should we be completely avoiding `NewAggregate` for now? Currently using `NewAggregate` in `CheckAndSetDefaults`, `updateStats`, and `verifyNethealth`. Should these all be replaced to fail fast and return individual errors? 

If we aggregate the errors for `verifyNethealth` we can return a single probe indicating all the nodes that it is unable to communicate with:
```bsh
* dev-bk-gravity-node-0 / 10.128.0.70 / node
            Status:     degraded
            [×]         failed nethealth test (network communication failure with 10.128.0.96, network communication failure with 10.128.0.97)
        * dev-bk-gravity-node-1 / 10.128.0.97 / node
            Status:     degraded
            [×]         failed nethealth test (network communication failure with 10.128.0.70, network communication failure with 10.128.0.96)
        * dev-bk-gravity-node-2 / 10.128.0.96 / node
            Status:     degraded
            [×]         failed nethealth test (network communication failure with 10.128.0.70, network communication failure with 10.128.0.97)
```

without aggregating errors we can return multiple probes, one for each peer:
```bsh
        * dev-bk-gravity-node-0 / 10.128.0.70 / node
            Status:     degraded
            [×]         nethealth-checker (network communication failure with 10.128.0.96)
            [×]         nethealth-checker (network communication failure with 10.128.0.97)
        * dev-bk-gravity-node-1 / 10.128.0.97 / node
            Status:     degraded
            [×]         nethealth-checker (network communication failure with 10.128.0.70)
            [×]         nethealth-checker (network communication failure with 10.128.0.96)
        * dev-bk-gravity-node-2 / 10.128.0.96 / node
            Status:     degraded
            [×]         nethealth-checker (network communication failure with 10.128.0.70)
            [×]         nethealth-checker (network communication failure with 10.128.0.97)
```

The second output looks a bit easier to read. Is this format more consistent with how were reporting failed probes?

bernardjkim:
With the current implementation, the probes do not provide much insight into why the nethealth checker is failing.

Example:
If we have a 3 node cluster (node-0, node-1, node2) and add a firewall rule to node-1 to drop udp traffic from node-0 we will end up with:
```bsh
* dev-bk-gravity-node-0 / 10.128.0.70 / node
    Status:     degraded
    [×]         nethealth-checker (network communication failure with 10.128.0.97)
* dev-bk-gravity-node-1 / 10.128.0.97 / node
    Status:     degraded
    [×]         nethealth-checker (network communication failure with 10.128.0.70)
* dev-bk-gravity-node-2 / 10.128.0.96 / node
    Status:     healthy
```
The failed probes indicate that node-0 and node-1 are failing to communicate, but doesn't really have any indication why.

Should we add additional steps to this checker? Once we detect that the overlay network is unhealthy, should we run additional tests to diagnose common problems with the network. For example, checking iptables? Not sure what other common problems users have. 

@r0mant, @knisbet Is this worth looking into or outside the scope of this feature?

r0mant:
I think this is good for now. There is a clear indication that there is a network communication problem b/w these 2 nodes which is a good kick-start for troubleshooting - and there may be myriads of reasons. It would certainly be helpful to be able to detect the most common ones, but I don't think we should be spending time on this right now.

monitoring/nethealth.go
```diff
@@ -0,0 +1,321 @@
...
+		}
+
+		series, err := c.getTimeoutSeries(peerName)
+		if err != nil {
+			errors = append(errors, trace.Wrap(err))
+			continue
+		}
+
+		// Keep only the last `seriesCapacity` number of data points.
+		if len(series) >= c.SeriesCapacity && len(series) > 0 {
```
r0mant: Nit: I think `c.SeriesCapacity` is always greater than 0 so the second `&& len(series) > 0` check isn't necessary here.

monitoring/nethealth.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	return metrics, nil
+}
+
+// updateStats updates the timeout data with new data points collected in
+// the provided metrics. Returns a list containing the updated keys.
+func (c *nethealthChecker) updateStats(metrics []*dto.Metric) (updated []string, err error) {
+	var errors []error
+	for _, metric := range metrics {
```
r0mant: So what are "metrics" here? Is it a single datapoint per peer representing whether the last nethealth check was success/failure? Or are these multiple binary (1/0) datapoints for each peer? Or number of timeouts for each peer? Could you describe the data model somewhere in the comments please.

monitoring/nethealth.go
```diff
@@ -0,0 +1,321 @@
...
+// isHealthy returns false if the number of timeouts increases at each data point.
+func (c *nethealthChecker) isHealthy(series []int64) bool {
+	// Checker has not collected enough data yet to check network health.
+	if len(series) < c.SeriesCapacity {
+		return true
+	}
+
+	// Increase in number of timeouts at each data point indicates network issue.
+	for i := 0; i < c.SeriesCapacity-1; i++ {
+		if series[i] >= series[i+1] {
```
r0mant: Trying to understand this algo - this kinda goes to my question about the data model above, what does `series[i]` represent? Number of timeouts during previous interval?

If so, won't this algorithm fail to detect the following scenario for example:

`5,6,5,6,5,6,5,6,..`

i.e. the number of timeouts decreases and the checker will think it's healthy. Unless I'm missing something (will prob be more clear when you document the data model).

monitoring/nethealth.go
```diff
@@ -0,0 +1,321 @@
...
+	// Find nethealth pod with matching host ip address.
+	for _, pod := range pods.Items {
+		if pod.GetLabels()[nethealthLabel] != nethealthValue {
+			continue
+		}
+		if pod.Status.HostIP == c.HostIP {
+			return fmt.Sprintf("http://%s:%d", pod.Status.PodIP, c.NethealthPort), nil
+		}
+	}
+	return "", trace.NotFound("unable to find local nethealth pod")
```
r0mant: Scenario where nethealth pod can't be found should be handled gracefully. Think about the upgrades - a new planet agent comes up but monitoring app still hasn't been upgraded and as such there's no nethealth pod running yet.

I would handle this not found error on the caller side gracefully and maybe just log it but not fail the checker.

monitoring/nethealth.go
```diff
@@ -0,0 +1,321 @@
...
+// nethealthChecker checks network communication between peers.
+type nethealthChecker struct {
+	// NethealthConfig contains caller specified nethealth checker configuration
+	// values.
+	NethealthConfig
+	// lock access to timeoutStats
+	sync.Mutex
+	// timeoutStats maps a peer to a time series data interval containing the
+	// number of echo timeouts received by the specific peer for the set interval.
+	timeoutStats *holster.TTLMap
```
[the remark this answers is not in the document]
bernardjkim: I saw that previous usage of `gravitational/ttlmap` was replaced in https://github.com/gravitational/satellite/pull/163. I haven't really looked at the differences, but I just figured we're moving from `ttlmap` to `holster`.

r0mant (commented):
I think the algo needs some more tweaking to be more reliable/predictable, see my comment.

monitoring/nethealth.go
```diff
@@ -0,0 +1,340 @@
...
+// Implements health.Checker
+func (c *nethealthChecker) Name() string {
+	return nethealthCheckerID
+}
+
+// Check verifies the network is healthy.
+// Implements health.Checker
+func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter) {
+	if err := c.check(ctx, reporter); err != nil {
+		log.WithError(err).Error("Unable to verify nethealth.")
```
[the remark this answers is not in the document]
bernardjkim: c234ed6
Failure to get k8s pods or fetch metrics from nethealth pod will now report failed probes as well.

monitoring/nethealth.go
```diff
@@ -0,0 +1,340 @@
...
+	if len(series) < c.SeriesCapacity {
+		return true
+	}
+
+	// series contains time series data of the running total of timeouts for a peer.
+	// If the counter is increasing at each data point, that means requests to
+	// this peer have been timing consistently throughout this interval. This
+	// should indicate that there is a network issue.
+	for i := 0; i < c.SeriesCapacity-1; i++ {
+		if series[i] >= series[i+1] {
```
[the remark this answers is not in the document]
bernardjkim: 6e2409c
Determining nethealth using packet loss percentage should definitely be more reliable. This commit adds a default 20% packet loss threshold and a 5 minute interval. The interval is used along the the agent's status update timeout to calculate the number of data points to store in a series.

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+// nethealthChecker checks network communication between peers.
+type nethealthChecker struct {
+	// NethealthConfig contains caller specified nethealth checker configuration
+	// values.
+	NethealthConfig
+	// seriesCapacity specifies the number of data points to store in a series.
+	seriesCapacity int
+	// lock access to timeoutStats
```
r0mant: timeoutStats -> peerStats?

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+
+// Check verifies the network is healthy.
+// Implements health.Checker
+func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter) {
+	probes, err := c.check(ctx)
+	if err != nil {
+		log.WithError(err).Error("Unable to verify nethealth.")
+	}
+
+	if probes.NumProbes() != 0 {
```
r0mant: To be honest, instead of returning probes here, I would still pass reporter to check() function like it was before, but just made sure that check() returns "nil" in case nethealth pod wasn't found. This way it would plug seamlessly into the old logic and it would be cleaner because probes would only be created in a single place. For example:

```go
func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter) {
  probes, err := c.check(ctx)
  if err != nil {
    // add error probe
  }
  // add success probe
}

  addr, err := c.getNethealthAddr(pods)
  if err != nil && !trace.IsNotFound(err) {
    return trace.Wrap(err) // we got unexpected error, maybe networking-related, will add error probe above
  }
  if trace.IsNotFound(err) {
    log.Debug("Nethealth pod not found.")
    return nil // pod wasn't found, log and treat gracefully
  }
  //...
}

func (c *nethealthChecker) getNethealthAddr() (string, error) {
  pods, err := c.Client.CoreV1().Pods(nethealthNamespace).List(metav1.ListOptions{})
  if err != nil {
    return "", rigging.ConvertError(err) // this will convert error to a proper trace error, e.g. trace.NotFound
  }
  // ...
}
```

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+		health.AddFrom(reporter, probes)
+		return
+	}
+	reporter.Add(NewSuccessProbe(c.Name()))
+}
+
+func (c *nethealthChecker) check(ctx context.Context) (health.Reporter, error) {
+	reporter := new(health.Probes)
+	pods, err := c.Client.CoreV1().Pods(nethealthNamespace).List(metav1.ListOptions{})
```
r0mant: Like in my comment above, this should be encapsulated into getNethealthAddr.

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+	if err != nil {
+		reporter.Add(NewProbeFromErr(c.Name(), "failed to retrieve k8s pods", err))
```
r0mant: And these will go away as we'll only be adding probes in the top-level Check method.

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+		return reporter, trace.Wrap(err)
+	}
+
+	addr, err := c.getNethealthAddr(pods)
+	if err != nil {
+		return reporter, trace.Wrap(err, "failed to get local nethealth address")
+	}
+
+	mf, err := fetchMetrics(ctx, addr)
```
r0mant: Nit: What's "mf"?  Can we pick a better name for this variable?

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+// fetchMetrics collects the network metrics from the nethealth pod.
+func fetchMetrics(ctx context.Context, addr string) (metricFamilies map[string]*dto.MetricFamily, err error) {
+	client, err := roundtrip.NewClient(addr, "")
+	if err != nil {
+		return nil, trace.Wrap(err, "failed to connect to nethealth service at %s.", addr)
+	}
+
+	resp, err := client.Get(ctx, client.Endpoint("metrics"), url.Values{})
+	if err != nil {
+		return nil, trace.ConvertSystemError(err)
```
r0mant: Should be just trace.Wrap() I think, ConvertSystemError is usually used to convert Go errors from system methods like open file, etc.

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+		return nil, trace.Wrap(err)
+	}
+
+	return metricsFamilies, nil
+}
+
+// updateStats updates netStats with new data collected in the provided
+// MetricFamily.
+func (c *nethealthChecker) updateStats(echoRequests, echoTimeouts map[string]float64) (updated []string, err error) {
+	if len(echoRequests) != len(echoTimeouts) {
```
r0mant: Do these lengths have to match? What do these maps map from->to?

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+		// packet loss percentages this round.
+		if requestInc <= 0 {
+			log.Warn("Request counter did not increase. Nethealth pod may have restarted.")
+			continue
+		}
+
+		// It should not be possible for the timeout counter to have increased
+		// more than the request counter. Log and ignore this situation.
+		if timeoutInc > requestInc {
+			log.WithField("request inc", requestInc).
```
r0mant: Nit: We usually make log fields w/o spaces, "request-increase", "timeout-increase".

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+		// Calcuate counter increase since last check and replace previous total
+		requestInc := requests - data.prevRequestTotal
+		data.prevRequestTotal = requests
+		timeoutInc := timeouts - data.prevTimeoutTotal
+		data.prevTimeoutTotal = timeouts
+
+		// Request counter should be constantly increasing, if this is not the
+		// case, nethealth pod most likely reset the counter. Do no update
+		// packet loss percentages this round.
+		if requestInc <= 0 {
```
r0mant: Hm, so if nethealth pod restarts, this will keep being < 0 until the previously reached number of total requests is reached again, or am I missing anything? Should we reset the peer stats in this case also?

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+				WithField("timeout inc", timeoutInc).
+				Warn("Timeout counter increased more than request counter.")
+			continue
+		}
+
+		// Record the new packet loss percentage, but only keep the last `seriesCapacity` number of data points.
+		if len(data.packetLossPercentages) >= c.seriesCapacity {
+			data.packetLossPercentages = data.packetLossPercentages[1:]
+		}
+		packetLoss := timeoutInc / requestInc
```
r0mant: Interesting, so you're recording the packet loss after each interval and then compare each of them with the threshold below. Do you know how many datapoints nethealth produces b/w the checker runs? I'm just thinking, if it checks it once every 30 seconds, we may have not enough "increases" here and the packet loss will always be either close to 0 or to a 100. Have you considered to instead just accumulating these increases and then calculating packet loss once in "isHealthy" function - so we'd get the packet loss for the entire interval.

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+// getPeerName extracts the 'peer_name' value from the provided labels.
+func getPeerName(labels []*dto.LabelPair) (peer string, err error) {
+	for _, label := range labels {
+		if peerLabel == label.GetName() {
+			return label.GetValue(), nil
+		}
+	}
+	return "", trace.NotFound("unable to find required peer label")
```
r0mant: Error could be a bit better, like which label?

monitoring/nethealth.go
```diff
@@ -0,0 +1,411 @@
...
+		errors = append(errors, trace.BadParameter("host advertise ip must be provided"))
+	}
+	if c.NethealthPort == 0 {
+		c.NethealthPort = defaultNethealthPort
+	}
+	if c.NetStatsInterval == time.Duration(0) {
+		c.NetStatsInterval = defaultNetStatsInterval
+	}
+	if c.KubeConfig == nil {
+		errors = append(errors, trace.BadParameter("kubernetes access config must be provided"))
```
a-palchikov: nit: move the test before the defaults are set.

monitoring/nethealth.go
bernardjkim: Modified to print out `peer_name` label.

monitoring/nethealth.go
bernardjkim: Ah ok, reverting back to previous structure. 
Currently unable to import a compatible version of `rigging`. Opened a PR to bump k8s and set a trace version https://github.com/gravitational/rigging/pull/75

r0mant: Hmm, since we need this in 5.5 I'm not sure we can safely bump Kubernetes version in rigging and revendor here. Might run into incompatibility issues - 5.5 has Kubernetes 1.13.

r0mant: @bernardjkim Btw on that note - I have just noticed that this PR is open to master. Shouldn't it be pointing to version/5.5.x instead? Because we're essentially backporting all this to 5.5 so all code changes should go to version/5.5.x branches of respective codebases (satellite, planet, etc.)

monitoring/nethealth.go
bernardjkim: They should match. The metrics should come in a pair (requestTotal, timeoutTotal). I've just wrapped both into a `networkData` struct and have `parseMetrics` return `networkData`

monitoring/nethealth.go
bernardjkim: Ah you're right. The stored previous totals should be updated here.

monitoring/nethealth.go
bernardjkim: After thinking about this some more, I think calculating the overall packet loss once for the interval may give some undesired results.

Example:
Two nodes are unable to communicate on the overlay network for some time and end up with 100% packet loss for the entire collection interval. When the overlay network issue is solved and the two nodes are able to communicate, the checker will continue to be fail until the overall packet loss percentage has decreased below the threshold. If starting from 100% this will probably require no packet loss for the entire collection interval.

monitoring/nethealth_test.go
```diff
@@ -0,0 +1,211 @@
...
+				timeoutInc: []float64{10, 20, 30, 40, 0},
+			},
+		},
+		{
+			comment:  Commentf("Expected failed probe. Timeouts increase at each interval."),
+			expected: &health.Probes{NewProbeFromErr(nethealthCheckerID, nethealthDetail(testNode), nil)},
+			storedData: peerData{
+				requestInc: []float64{10, 20, 30, 40, 50},
+				timeoutInc: []float64{10, 20, 30, 40, 50},
+			},
```
a-palchikov: You don't exercise the length difference test between requests/timeouts which is handled in code.

monitoring/nethealth.go
```diff
@@ -0,0 +1,448 @@
...
+	if err != nil {
+		return trace.Wrap(err, "failed to update nethealth stats")
+	}
+
+	return c.verifyNethealth(updated, reporter)
+}
+
+// getNethealthAddr returns the address of the local nethealth pod.
+func (c *nethealthChecker) getNethealthAddr() (addr string, err error) {
+	pods, err := c.Client.CoreV1().Pods(nethealthNamespace).List(metav1.ListOptions{})
```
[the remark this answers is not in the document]
bernardjkim: Oh, did not realize we could filter the pods with `ListOptions`, but that makes sense that functionality would be provided by k8s api. I'll try to be more thorough when learning to use a new api.

monitoring/nethealth.go
```diff
@@ -0,0 +1,448 @@
...
+	// Calcuate counter increase since last check and replace previous total
+	requestInc := incomingData.requestTotal - storedData.prevRequestTotal
+	timeoutInc := incomingData.timeoutTotal - storedData.prevTimeoutTotal
+	storedData.prevRequestTotal = incomingData.requestTotal
+	storedData.prevTimeoutTotal = incomingData.timeoutTotal
+
+	// Request counter should be constantly increasing, if this is not the
+	// case, nethealth pod most likely reset the counter. Update prevTotals,
+	// but do no update requestInc/timeoutInc.
+	if requestInc <= 0 {
```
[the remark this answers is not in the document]
bernardjkim: Yes, `timeoutInc` could also be negative if the nethealth pod restarts. I'll add an additional check.

monitoring/nethealth.go
```diff
@@ -0,0 +1,448 @@
...
+// isHealthy returns true if the overlay network is healthy for the specified
+// peer.
+func (c *nethealthChecker) isHealthy(peer string) (healthy bool, err error) {
+	storedData, err := c.peerStats.Get(peer)
+	if err != nil {
+		return false, trace.Wrap(err)
+	}
+
+	// Checker has not collected enough data yet to check network health.
+	if len(storedData.requestInc) < c.seriesCapacity {
```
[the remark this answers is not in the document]
bernardjkim: 2365796
Modified `netStats` and can now be initialized with a `packetLossCapacity`. `packetLoss` will be initialized with a constant capacity.

monitoring/nethealth.go
```diff
@@ -0,0 +1,448 @@
...
+	// echoTimeouts maps a peer to the current running total number of timeouts received from that peer.
+	echoTimeouts, err := parseCounter(metricFamilies, echoTimeoutLabel)
+	if err != nil {
+		return nil, trace.Wrap(err, "failed to parse echo timeouts")
+	}
+
+	// Each peer should have a request counter and timeout counter pair.
+	// If lengths are unequal, we are missing data.
+	if len(echoRequests) != len(echoTimeouts) {
+		return nil, trace.BadParameter("recieved %d timeout counters and %d request counters",
```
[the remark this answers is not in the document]
bernardjkim: 2f089b7
Moved parsing functionality into fetchMetrics. If we fail to parse metrics we will now log the metrics in text format received from nethealth. Also added an example of expected nethealth metrics. 

We will only run into this situation if there is an bug in nethealth.

monitoring/nethealth_test.go
bernardjkim: 1c9ec8b
Added some test cases for parseMetrics.

monitoring/nethealth.go
```diff
@@ -203,17 +204,19 @@ func (c *nethealthChecker) updatePeer(peer string, incomingData networkData) err
 		return trace.Wrap(err)
 	}
 
-	// Calcuate counter increase since last check and replace previous total
+	// Calcluate counter delta since last check and replace total.
```
a-palchikov: Cal**cul**ate :-)

monitoring/nethealth.go
```diff
@@ -203,17 +204,19 @@ func (c *nethealthChecker) updatePeer(peer string, incomingData networkData) err
...
 	storedData.prevTimeoutTotal = incomingData.timeoutTotal
 
-	// Request counter should be constantly increasing, if this is not the
-	// case, nethealth pod most likely reset the counter. Update prevTotals,
-	// but do no update requestInc/timeoutInc.
-	if requestInc <= 0 {
-		log.Warn("Request counter did not increase. Nethealth pod may have restarted.")
+	// Request counter should be strictly increasing and timeout counter should
+	// be monotonically increasing. If this is not the case, nethealth pod most
+	// likely restart and reset the counters.
```
a-palchikov: restart**ed** and reset the counters

r0mant (approved):

PR #165, opened 2020-03-04
[5.5.x] add profiling to the HTTPS health endpoint

a-palchikov:
This PR adds profiling endpoints on the shared health endpoint behind mTLS.

Updates https://github.com/gravitational/gravity/issues/1146 and https://github.com/gravitational/gravity/issues/1091.


knisbet:
Oops, left the same comment on the planet side of the PR. 

Even though it's behind mTLS, I'm not sure if leaving a debug/profiling endpoint available would be considered a best practice. My suspicion would be that it could be flagged in an audit, despite being behind mTLS, we give every node and tsh login a set of TLS keys which I believe can reach the particular endpoint.

r0mant (approved):
Code looks good.

@knisbet Are there any specific/official guidelines that discourage having profiling endpoints enabled? I was under the impression that whenever these were highlighted during security scans, it was because they were unprotected?

bernardjkim (approved):

a-palchikov:
@r0mant @knisbet I believe Kevin is right - this needs to be properly isolated and ideally on demand - only when troubleshooting.

Options:
  * only expose it on the localhost
  * offer only the goroutine dump endpoint without exposing the rest of the pprof
  * offer debugging endpoint on demand

a-palchikov:
I implemented the hybrid approach:
  * debug endpoints are only exposed on localhost
  * debug endpoints must be explicitly enabled to use
  * debug endpoints can be turned off when done

bernardjkim:
I think the code looks good :+1: 

I was looking into an alternate approach I'd like to suggest if it hasn't already been considered. Instead of serving profiles on a server, we could provide a command to save the profiles to a file. We could have something like `planet profile cpu` that saves the cpu profile to a file. We can provide additional subcomands for the different types of profiles and a flags to specify file location or maybe profiling duration.

For the code we could use https://github.com/pkg/profile and a profiling function might look something like: 
```golang
func collectCPUProfile(filename string) {
	defer profile.Start(profile.CPUProfile, profile.ProfilePath(filename)).Stop()
	clock := clockwork.NewRealClock()
	ticker := clock.NewTicker(5 * time.Second)
	<-ticker.Chan()
}
```

The benefit of this approach is a simpler ux. Since the profiling endpoint needs to be toggled on/off, I'm wondering if users may forget to disable the profiling endpoint once they are done debugging.

a-palchikov:
@bernardjkim @knisbet @r0mant I've removed the debug endpoint and instead added a new debug service (which is atm also included with the agent RPC interface but can be definitely used separately).
The idea is to only have access to it in the local context (i.e. from within the container). To move it in that direction, satellite could have the server only available on a unix socket, for example.

agent/proto/debug/debug.go
```diff
@@ -0,0 +1,68 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// NewServer creates a new instance of the Debug service
+func NewServer() *Server {
+	return &Server{}
+}
+
+// Profile executes the specifies debug profile and streams the results to the caller
```
bernardjkim, line 31: nit: specifies -> specified

lib/rpc/net.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+	"net"
+
+	"github.com/gravitational/trace"
+)
+
+// SplitHostPort splits the specified host:port into host/port parts.
+// If the input address does not have a port, defaultPort as appended.
+// Returns the extracted IP address and port
```
bernardjkim, line 28: nit: The godocs make it sound like two return values are expected, an address and a port number. Could be reworded to clarify.

nit: defaultPort ~as~ is appended

r0mant (approved):

agent/agent.go
```diff
@@ -156,18 +174,13 @@ type agent struct {
...
-	// timeline is only used by members that have the role 'master'.
-	ClusterTimeline history.Timeline
-
-	// LocalTimeline keeps track of local timeline events.
-	LocalTimeline history.Timeline
+	// cancel is used to cancel the internal processes
+	// running as part of g
+	cancel context.CancelFunc
+	// ctx manages the lifetime of the agent's internal processes
+	ctx context.Context
```
knisbet, line 181: nit: I know it makes it easier, but I thought the idiomatic pattern for context is that it should not be placed within a struct?

build.go
```diff
@@ -176,7 +175,7 @@ func (Test) Unit() error {
 	mg.Deps(Build.BuildContainer)
 	fmt.Println("\n=====> Running Satellite Unit Tests...\n")
 	return trace.Wrap(sh.RunV(
-		"docker", "run", "-it", "--rm=true",
+		"docker", "run", "-t", "--rm",
```
knisbet, line 178: I think -t can probably be removed as well. I definitely originally wrote this from muscle memory of running docker on the cli, and I think it does cause some weird behaviour indicating there is a tty attached when there isn't actually one.

lib/rpc/net.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"net"
+
+	"github.com/gravitational/trace"
+)
+
+// SplitHostPort splits the specified host:port into host/port parts.
+// If the input address does not have a port, defaultPort is appended.
+// Returns the extracted IP address and port as a net.TCPAddr
+func SplitHostPort(hostPort string, defaultPort int) (addr *net.TCPAddr, err error) {
```
knisbet, line 29: nit: function name doesn't seem to match use. IE this function doesn't split a string into host and port, instead it seems to resolveTCPAddr with the added ability of specifying a default port if unspecified.

knisbet (approved):

agent/agent.go
a-palchikov: It is not always possible to do that - in this case, since I went with `errgroup`, and additionally wanted one central group for the agent. Now I'm not able to access the original context the group was created with since the functions I pass to `errgroup.Go` do not accept any context and the errgroup API does not provide access to the underlying context - which is unfortunate.
`errgroup` seems to be the go-to package w.r.t handling groups of goroutines so I consider it an upgrade.
[Here](https://github.com/golang/go/issues/34510) is a relevant discussion about this shortcoming which is still unresolved.

PR #167, opened 2020-03-11
Clean up & add LeaderElected event

bernardjkim:
Refactor/Cleanup:
- Renamed agent `Timeline` to `ClusterTimeline`. Easier to differentiate between local and cluster timelines.
- Moved protobuf event constructors into `agentpb` package from the `history` package.
- Update `Timeline` interface. Renamed `RecordTimeline` to `RecordEvents`. Removed `RecordStatus`. The timeline does not need to be responsible for keeping track of the cluster status. Agent already keeps track of this.
- Updated MemTimeline to utilize the interfaces `DataInserter`, `Execer`.
- Moved some code around.

Changes required to record changes to cluster leadership:
- Added new `LeaderElected` protobuf message.
- Added `RecordLocalEvents` to `Agent` interface. This will allow planet to manually record events when a new leader is elected. Currently, the local timeline is only updated automatically during status update loop.


lib/history/memory/inmemory.go
```diff
@@ -37,155 +36,106 @@ import (
 //
 // Implements history.Timeline
 type Timeline struct {
-	sync.Mutex
 	// clock is used to record event timestamps.
 	clock clockwork.Clock
 	// capacity specifies the max number of events stored in the timeline.
 	capacity int
+	// lock access to events
```
r0mant: Nit: "Mutex locks access to events."

lib/history/sqlite/event.go
```diff
@@ -106,240 +77,184 @@ type sqlExecer struct {
...
+	*pb.TimelineEvent
 }
 
-// Insert inserts event into storage using the provided execer.
-func (e nodeRemoved) Insert(ctx context.Context, execer history.Execer) error {
+func (r *nodeRemoved) Insert(ctx context.Context, execer history.Execer) error {
 	const insertStmt = "INSERT INTO events (timestamp, type, node) VALUES (?,?,?)"
-	args := []interface{}{e.Timestamp, e.EventType, e.Node.String}
+	e := r.GetNodeRemoved()
+	if e == nil {
```
r0mant: This is also not very clear, what's GetNodeRemoved() supposed to return, why can it be nil and what to do in this case? Sorry, commenting on existing code but just thinking this may be confusing if anybody else needs to add a new timeline event for example.

lib/history/sqlite/sqlite.go
```diff
@@ -196,13 +165,36 @@ func (t *Timeline) RecordTimeline(ctx context.Context, events []*pb.TimelineEven
...
+	for _, event := range events {
+		row, err := newDataInserter(event)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+
+		err = row.Insert(ctx, sqlExecer)
+		// Unique constraint error indicates duplicate row.
+		// Just ignore duplicates and continue.
+		if isErrConstraintUnique(err) {
```
r0mant: Should we log it?

lib/history/sqlite/sqlite.go
```diff
@@ -196,13 +165,36 @@ func (t *Timeline) RecordTimeline(ctx context.Context, events []*pb.TimelineEven
...
+			continue
+		}
+		if err != nil {
+			return trace.Wrap(err)
```
r0mant: Should this method attempt to insert all events rather than bailing on first error? Or are they gonna be retried later?

agent/proto/agentpb/event.go
```diff
@@ -0,0 +1,123 @@
...
+		},
+	}
+}
+
+// NewClusterRecovered constructs a new ClusterRecovered event with the
+// provided data.
+func NewClusterRecovered(timestamp time.Time) *TimelineEvent {
+	event := newTimelineEvent(timestamp)
+	event.Data = &TimelineEvent_ClusterRecovered{
+		ClusterRecovered: &ClusterRecovered{},
```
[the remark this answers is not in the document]
bernardjkim: f113f78
Renamed `ClusterRecovered` -> `ClusterHealthy` and `NodeRecovered` -> `NodeHealthy`
Requires update to gravity's `printEvent` function https://github.com/gravitational/gravity/blob/3041f5adea706cf0601cd5268076b8539aa6d7b9/tool/gravity/cli/history.go#L61

lib/history/memory/event.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func (r *nodeRemoved) Insert(ctx context.Context, execer history.Execer) error {
+	const insertStmt = "timestamp,type,node"
+	e := r.GetNodeRemoved()
+	if e == nil {
+		return trace.BadParameter("expected NodeRemoved, got nil")
+	}
+	args := []interface{}{r.GetTimestamp().ToTime(), history.NodeRemoved, e.GetNode()}
+	return trace.Wrap(execer.Exec(ctx, insertStmt, args...))
```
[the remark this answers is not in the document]
bernardjkim: Execer interface takes the place of a db. `MemTimeline` just stores events in an array so the interface might not really work here, but it makes more sense when using sqlite or other db. 

:sweat_smile: packing the fields into a variable first was my attempt at writing cleaner code. How about we just move that portion into a separate `args` method to return the list of arguments.

lib/history/sqlite/event.go
bernardjkim: d65aa55
`Get<EventType>` does a type assertion and returns the event if ok, or nil o.w.
Replaced the function call and just use a type assertion instead. Should be easier to infer functionality of the code.

lib/history/sqlite/sqlite.go
bernardjkim:
```golang
// UpdateTimeline updates the timeline with a new event.
// Duplicate requests will have no effect.
func (r *server) UpdateTimeline(ctx context.Context, req *pb.UpdateRequest) (*pb.UpdateResponse, error) {
	if err := r.agent.RecordTimeline(ctx, []*pb.TimelineEvent{req.GetEvent()}); err != nil {
		return nil, GRPCError(err)
	}
	if err := r.agent.RecordLastSeen(req.GetName(), req.GetEvent().GetTimestamp().ToTime()); err != nil {
		return nil, GRPCError(err)
	}
	return &pb.UpdateResponse{}, nil
}
```
If the agent fails to record all events, the last seen timestamp will not be updated and they will be retried next time.
Maybe `RecordTimeline` should at least return the timestamp of the last event it recorded.

lib/history/memory/event.go
```diff
@@ -187,11 +187,12 @@ type nodeAdded struct {
 
 func (r *nodeAdded) Insert(ctx context.Context, execer history.Execer) error {
 	const insertStmt = "timestamp,type,node"
-	e := r.GetNodeAdded()
-	if e == nil {
-		return trace.BadParameter("expected NodeAdded, got nil")
+	data, ok := r.GetData().(*pb.TimelineEvent_NodeAdded)
```
a-palchikov: Not sure why you would keep the boxed `Data` inside specific event types:
```go
// newDataInserter returns the event as a history.DataInserter.
func newDataInserter(event *pb.TimelineEvent) (row history.DataInserter, err error) {
	switch t := event.GetData().(type) {
	case *pb.TimelineEvent_ClusterDegraded:
		return &clusterDegraded{ts: event.Timestamp}, nil
	case *pb.TimelineEvent_ClusterHealthy:
		return &clusterHealthy{ts: event.Timestamp}, nil
	case *pb.TimelineEvent_NodeAdded:
		return &nodeAdded{ts: event.Timestamp, data: t.NodeAdded}, nil
	case *pb.TimelineEvent_NodeRemoved:
		return &nodeRemoved{ts: event.Timestamp, data: t.NodeRemoved}, nil
	case *pb.TimelineEvent_NodeHealthy:
		return &nodeHealthy{ts: event.Timestamp, data: t.NodeHealthy}, nil
	case *pb.TimelineEvent_NodeDegraded:
		return &nodeDegraded{ts: event.Timestamp, data: t.NodeDegraded}, nil
	case *pb.TimelineEvent_ProbeSucceeded:
		return &probeSucceeded{ts: event.Timestamp, data: t.ProbeSucceeded}, nil
	case *pb.TimelineEvent_ProbeFailed:
		return &probeFailed{ts: event.Timestamp, data: t.ProbeFailed}, nil
	case *pb.TimelineEvent_LeaderElected:
		return &leaderElected{ts: event.Timestamp, data: t.LeaderElected}, nil
	default:
		return row, trace.BadParameter("unknown event type %T", t)
	}
}
// ...
// clusterDegraded represents a cluster degraded event.
//
// Implements history.DataInserter.
type clusterDegraded struct {
	ts *pb.Timestamp
}
// ...
// nodeAdded represents a node added event.
//
// Implements history.DataInserter.
type nodeAdded struct {
	ts   *pb.Timestamp
	data *pb.NodeAdded
}
// ...
func (r *nodeAdded) Insert(ctx context.Context, execer history.Execer) error {
	const insertStmt = "timestamp,type,node"
	return trace.Wrap(execer.Exec(ctx, insertStmt, r.ts.ToTime(), history.NodeAdded, r.data.Node))
}
```
Likewise for the rest of the event types.

lib/history/memory/event.go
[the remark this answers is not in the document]
a-palchikov: See my suggestion above - there's no need to pack them into a slice of interfaces prior to calling into `Exec`.

agent/agent.go
```diff
@@ -449,19 +449,25 @@ func (r *agent) runChecks(ctx context.Context) *pb.NodeStatus {
...
-		return r.Timeline.GetEvents(ctx, params)
+		return r.ClusterTimeline.GetEvents(ctx, params)
 	}
 	return nil, trace.BadParameter("requesting cluster timeline from non master")
 }
 
-// RecordTimeline records the events into the cluster timeline.
-func (r *agent) RecordTimeline(ctx context.Context, events []*pb.TimelineEvent) error {
+// RecordClusterEvents records the events into the cluster timeline.
+// Cluster timeline can only be updated if agent has role 'master'.
```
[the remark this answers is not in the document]
bernardjkim: This might be a good idea since Agent is starting to look a little bloated. Not sure how big a change this will require. Will probably make changes in a separate PR.

lib/history/memory/event.go
bernardjkim: Ah this makes more sense.
Made changes in 8d81ddf

agent/proto/agentpb/agent.proto
```diff
@@ -167,21 +167,23 @@ message TimelineEvent {
...
     
-    NodeRecovered nodeRecovered = 7;
+    NodeHealthy nodeHealthy = 7;
     NodeDegraded nodeDegraded = 8;
     
     ProbeSucceeded probeSucceeded = 9;
     ProbeFailed probeFailed = 10;
 
-    // TODO: Additional events
+    LeaderElected leaderElected = 11;
```
r0mant, line 184: Really a nit, I would add godocs on these exported fields so they're populated for the generated structs too.

lib/history/event.go
```diff
@@ -38,97 +37,32 @@ type Execer interface {
...
-			Node:  node,
-			Probe: probe,
-		},
-	}
-	return event
-}
+// EventType specifies the type of event.
+type EventType string
+
+// Defines event types
```
r0mant, line 46: Nit: It's pretty clear as it is :)

lib/history/sqlite/sqlite.go
```diff
@@ -196,13 +165,38 @@ func (t *Timeline) RecordTimeline(ctx context.Context, events []*pb.TimelineEven
...
 	}
 
 	return nil
 }
 
+// insertEvents inserts the provided events into the timeline.
+func (t *Timeline) insertEvents(ctx context.Context, events []*pb.TimelineEvent) error {
+	sqlExecer := newSQLExecer(t.database)
+	for _, event := range events {
+		row, err := newDataInserter(event)
```
r0mant, line 179: Really a nit, but in scenarios like this when you iterate over something and process items one-by-one, I like to split the "single item processing" into a separate method as well which makes it a bit cleaner (and potentially, more reusable):

```go
func (t *Timeline) insertEvents(ctx context.Context, events []*pb.TimelineEvent) error {
  for _, event := range events {
    if err := t.insertEvent(ctx, event); err != nil {
      return trace.Wrap(err)
    }
  }
  return nil
}

func (t *Timeline) insertEvent(ctx context.Context, event *pb.TimelineEvent) error {
  // process event
}
```

r0mant (approved):

a-palchikov (approved):

PR #171, opened 2020-03-13
Fix a potential data race with TLS config and ListenAndServeTLS.

a-palchikov:
Earlier today I ran into a `SIGSEGV` with the agent which led me to this discovery.
The [ListenAndServeTLS](https://github.com/golang/go/blob/release-branch.go1.12/src/net/http/h2_bundle.go#L3834) has a path where it **modifies** the TLS configuration, so it's not safe to pass to multiple instances simultaneously.


bernardjkim (approved):

r0mant (approved):
This is a good one.

PR #172, opened 2020-03-18
[5.5.x] Add nethealth checker

bernardjkim:
- Backport nethealth #113.
- Add nethealth checker.


Gopkg.toml
```diff
@@ -57,7 +60,7 @@
   name = "github.com/gravitational/roundtrip"
 
 [[constraint]]
-  version = "1.1.7"
+  version = "1.1.8"
```
a-palchikov: I think it's safe and recommended to use the latest release (1.1.10) here (and in gravity) as it fixes a number of issues.

cmd/nethealth/main.go
```diff
@@ -0,0 +1,111 @@
...
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+func init() {
+	version.Init("v0.0.1-master+$Format:%h$")
+}
+
+func main() {
+	if err := run(); err != nil {
+		log.Fatal(trace.DebugReport(err))
```
a-palchikov: `log.WithError(err).Error("Failed.")` Otherwise the stack trace will be unreadable due to quoting.

cmd/nethealth/main.go
```diff
@@ -0,0 +1,111 @@
...
+		crunNodeName = crun.Flag("node-name", "The name of the node we're running on").
+				OverrideDefaultFromEnvar("NODE_NAME").String()
+		crunSelector = crun.Flag("pod-selector", "The kubernetes selector to identify nethealth pods").
+				Default(nethealth.DefaultSelector).String()
+	)
+
+	var cmd string
+	var err error
+
+	cmd, err = app.Parse(os.Args[1:])
```
a-palchikov: nit: technically, you don't need to pre-declare the  variables (unless you were initializing them conditionally):
```go
cmd, err := app.Parse(os.Args[1:])
```

cmd/nethealth/main.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return trace.Errorf("unable to parse command line.\nUse nethealth --help for help.")
```
a-palchikov: `trace.BadParameter`

cmd/nethealth/main.go
```diff
@@ -0,0 +1,111 @@
...
+			return trace.Wrap(err)
+		}
+
+		err = server.Start()
+		if err != nil {
+			return trace.Wrap(err)
+		}
+
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
```
a-palchikov: I'm wondering why the signal handling here? The default behavior of both SIGINT and SIGTERM is to terminate the process and the handler does not do anything special other than log the signal. Is there maybe a way to trigger graceful shutdown of the server?

monitoring/nethealth.go
```diff
@@ -0,0 +1,483 @@
...
+	}
+
+	// echoTimeouts maps a peer to the current running total number of requests that timed out.
+	echoTimeouts, err := parseCounter(metricFamilies, echoTimeoutLabel)
+	if err != nil {
+		return nil, trace.Wrap(err, "failed to parse echo timeouts")
+	}
+
+	if len(echoRequests) != len(echoTimeouts) {
+		return nil, trace.BadParameter("expected equal number of counters for requests and timeouts,"+
```
a-palchikov: So I'm still uncertain about this error. Can you clarify how this can be resolved (the comment mentions it might be due to change of labels - does this mean invalid configuration or something else)?

monitoring/nethealth.go
```diff
@@ -0,0 +1,483 @@
...
+	// echoTimeoutLabel defines the metric family label for the echo timeout counter.
+	echoTimeoutLabel = "nethealth_echo_timeout_total"
+
+	// defaultNethealthPort defines the default nethealth port.
+	defaultNethealthPort = 9801
+
+	// defaultNetStatsInterval defines the default interval duration for the netStats.
+	defaultNetStatsInterval = 5 * time.Minute
+
+	// netStatsCapacity sets the number of TTLMaps that can be stored.
```
a-palchikov: Not sure I follow this comment. How is this affected by the cluster size being a const?
Also probably needs rephrasing:
```
// netStatsCapacity specifies the maximum number of nethealth samples to store.
```

monitoring/nethealth.go
```diff
@@ -0,0 +1,483 @@
...
+	// This will be the size of the cluster -1.
+	netStatsCapacity = 1000
+
+	// netStatsTTLSeconds defines the time to live in seconds for the stored
+	// netStats. This ensures the checker does not hold on to unused data when
+	// a member leaves the cluster.
+	netStatsTTLSeconds = 60 * 60 * 24 // 1 day
+
+	// packetLossThreshold defines the packet loss percentage used to determine
+	// if overlay network communication with a peer is unhealthy.
```
a-palchikov: Needs adding that the loss needs to be consistently observed over the entire interval and not on individual samples

a-palchikov (approved):
lgtm with a couple of comments

monitoring/nethealth.go
bernardjkim: Yeah, I'm still trying to figure out how to properly handle this. TTLMap needs to be initialized with a capacity that specifies the max number of elements it can store. This should be equal to the number of members in the cluster. However, the number of members in a cluster is dynamic since members can join/leave. 
Will specify this is just default max number of cluster members for now.

cmd/nethealth/main.go
bernardjkim: It doesn't look like nethealth currently provides any functionality to gracefully shutdown the server. I'd rather not dig into the nethealth code right now, but I'll take a look into this at a later time.

monitoring/nethealth.go
```diff
@@ -0,0 +1,484 @@
...
+}
+
+// Set maps the specified peer and data.
+func (r *netStats) Set(peer string, data peerData) error {
+	r.Lock()
+	defer r.Unlock()
+	return r.TTLMap.Set(peer, data, netStatsTTLSeconds)
+}
+
+// peerData keeps trock of relevant nethealth data for a node.
```
r0mant: track

r0mant (approved):
lgtm, I'd be curious to see how this algorithm (calculating timeouts over each interval vs over the entire sample array) performs in the field.

monitoring/nethealth.go
```diff
@@ -262,38 +263,59 @@ func (c *nethealthChecker) isHealthy(peer string) (healthy bool, err error) {
...
-	//      nethealth_echo_request_total{node_name="10.128.0.96",peer_name="10.128.0.70"} 236
-	//      nethealth_echo_request_total{node_name="10.128.0.96",peer_name="10.128.0.97"} 273
-	//      # HELP nethealth_echo_timeout_total The number of echo requests that have timed out
-	//      # TYPE nethealth_echo_timeout_total counter
-	//      nethealth_echo_timeout_total{node_name="10.128.0.96",peer_name="10.128.0.70"} 37
-	//      nethealth_echo_timeout_total{node_name="10.128.0.96",peer_name="10.128.0.97"} 0
-	resp, err := client.Get(ctx, client.Endpoint("metrics"), url.Values{})
+	respCh := make(chan *http.Response, 1)
+	errCh := make(chan error, 1)
+	go func() {
```
a-palchikov: I'm seeing this goroutine as an attempt to handle context expiration and HTTP query simultaneously - is this the case?
Why not use a request with a context for that case (as it sort of was before)? Also, using the default HTTP client is not ideal in most of the cases due to its implicit configuration w.r.t timeouts.

bernardjkim: Yeah, I wanted to handle an http request with context here without using the `roundtrip` package. 

Gravity 5.5 is using an older version of `roundtrip`. So if we wanted to use `roundtrip` here we would need to update gravity's version of the package and all of the code that imports the package.

I think gravity 6.x has updated the `roundtrip` package, so we can use the previous implementation in versions >= 6.x

a-palchikov: You don't need roundtrip to make an HTTP request bound to a context:
```go
// create a custom HTTP client
req := http.NewRequestWithContext(ctx)
resp, err := client.Do(req)
```

PR #178, opened 2020-03-31
[5.5.x] Close expired client connections

bernardjkim:
Close client connections for nodes that have left the cluster. 

Updates gravitational/gravity#1280

### DIFF
--- a/monitoring/timedrift.go
+++ b/monitoring/timedrift.go
@@ -29,7 +29,6 @@ import (
 	"github.com/gravitational/satellite/lib/rpc/client"
 
 	"github.com/gravitational/trace"
-	"github.com/gravitational/ttlmap"
 	serf "github.com/hashicorp/serf/client"
 	"github.com/jonboulle/clockwork"
 	log "github.com/sirupsen/logrus"
@@ -41,9 +40,6 @@ const (
 	// timeDriftThreshold sets the default threshold of the acceptable time
 	// difference between nodes.
 	timeDriftThreshold = 300 * time.Millisecond
-	// clientsCacheCapacity is the capacity of the TTL map that holds
-	// clients to satellite agents on other cluster nodes.
-	clientsCacheCapacity = 1000
 )
 
 // timeDriftChecker is a checker that verifies that the time difference between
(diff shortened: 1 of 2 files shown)


a-palchikov (approved):
This lgtm but I have doubts whether this could have been causing such leaks - was it the on a cluster where nodes leave frequently? If not, there must be another source of leaks

monitoring/timedrift.go
```diff
@@ -262,6 +257,28 @@ func (c *timeDriftChecker) nodesToCheck() (result []serf.Member, err error) {
...
+	defer c.mu.Unlock()
+
+	for addr, conn := range c.clients {
+		if _, ok := currentMembers[addr]; ok {
+			continue
+		}
+		if err := conn.Close(); err != nil {
+			log.WithError(err).Error("Failed to close client connection.")
+		}
+		delete(c.clients, addr)
```
r0mant, line 278: Add a log entry indicating that we closed this client?

monitoring/timedrift.go
```diff
@@ -262,6 +257,28 @@ func (c *timeDriftChecker) nodesToCheck() (result []serf.Member, err error) {
...
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	for addr, conn := range c.clients {
+		if _, ok := currentMembers[addr]; ok {
+			continue
+		}
+		if err := conn.Close(); err != nil {
+			log.WithError(err).Error("Failed to close client connection.")
```
r0mant, line 276: Add a log field with client address?

monitoring/timedrift.go
```diff
@@ -273,16 +290,18 @@ func (c *timeDriftChecker) shouldCheckNode(node serf.Member) bool {
 func (c *timeDriftChecker) getAgentClient(ctx context.Context, node serf.Member) (client.Client, error) {
 	c.mu.Lock()
```
r0mant, line 291: Sorry this is not a part of your changeset but I'm just noticing that we're acquiring a lock here over DialRPC function which actually attempts to dial a remote server. This can potentially lead to the entire checker stall if DialRPC hangs. Since you're here anyway, could you please rewrite this so we only place locks on map operations and not I/O? I.e. the connection should be created outside the lock.

r0mant (commented):
Agree with Dima, this is good to have but most certainly not the main source of leaks unless the cluster expands/shrinks with thousands of nodes.

bernardjkim:
After thinking about it some more, you guys are probably right. I don't think this could have leaked so many open connections. Will continue looking for other potential leaks.

PR #179, opened 2020-03-31
[5.5.x] Parsing error no longer tiggers failed probe

bernardjkim:
Parsing errors will only be logged and will no longer trigger a failed probe.

Resolves gravitational/gravity#1307

### DIFF
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -125,9 +125,17 @@ func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter) 
 		return trace.Wrap(err) // received unexpected error, maybe network-related, will add error probe above
 	}
 
-	netData, err := fetchNethealthMetrics(ctx, addr)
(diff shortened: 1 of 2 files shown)


r0mant (approved):
@bernardjkim Could you please also verify that clusters no longer enter degraded state due to this upon fresh install?

a-palchikov:
Why is there a parsing error in the first place - the referenced issue does not mention this?

bernardjkim:
@a-palchikov 
Good question, I added some additional information in the referenced issue. 

If you look through the received nethealth metrics, it does not contain any request/timeout counters. 
In order for nethealth to initialize its counters, it needs to first discover the other nethealth pods in the cluster, so there is some duration after initialization when nethealth metrics will return no counters.

This issue probably would have gone unnoticed since the initial discovery should run within ~10 seconds, but because I had removed the service resource from the monitoring-app, this initial discovery was unable to execute successfully.

PR #180, opened 2020-04-01
[5.5.x] Fix error handling

bernardjkim:
More error handling fixes.

- Log pod not found error with log level WARN instead of DEBUG.
- Report last recorded packet loss percentage when probes fails.
- Previously reporting additional success probe even if checks failed.

Also fix `getNethealthAddr` by removing the field selector. Cannot rely on `spec.nodeName` to be equal to the advertised IP. Also `status.hostIP` seems to be unsupported as a field selector.

### DIFF
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -164,7 +164,7 @@ func (c *nethealthChecker) getNethealthAddr() (addr string, err error) {
 		}
 	}
 
-	return addr, trace.NotFound("unable to find local nethealth pod")
+	return addr, trace.NotFound("unable to find nethealth pod running on host %s", c.AdvertiseIP)
 }
 
 // updateStats updates netStats with new incoming data.
@@ -248,6 +248,9 @@ func (c *nethealthChecker) verifyNethealth(peers []string, reporter health.Repor
 		if err != nil {
 			return trace.Wrap(err)
 		}
+		if len(data.packetLoss) == 0 {
+			continue
+		}
 		packetLoss := data.packetLoss[len(data.packetLoss)-1]
 		reporter.Add(nethealthFailureProbe(c.Name(), peer, packetLoss))
 	}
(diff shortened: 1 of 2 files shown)


monitoring/nethealth.go
```diff
@@ -260,6 +271,18 @@ func (c *nethealthChecker) isHealthy(peer string) (healthy bool, err error) {
 	return false, nil
 }
 
+// nethealthFailureProbe constructs a probe that represents failed nethealth check
+// against the specified peer.
+func nethealthFailureProbe(name, peer string, packetLoss float64) *pb.Probe {
+	return &pb.Probe{
+		Checker: name,
+		Detail: fmt.Sprintf("overlay packet loss for peer %s is higher than the allowed threshold of %.2f%%: %.2f%%",
+			peer, packetLossThreshold*100, packetLoss*100),
```
a-palchikov, line 280: nit: you could percompute threshold:
```go
const thresholdPercent = packetLossThreshold*100
```

monitoring/nethealth.go
```diff
@@ -228,9 +231,17 @@ func (c *nethealthChecker) verifyNethealth(peers []string, reporter health.Repor
...
+		if healthy {
+			continue
+		}
+
+		// Report last recorded packet loss percentage
+		data, err := c.peerStats.Get(peer)
+		if err != nil {
+			return trace.Wrap(err)
 		}
+		packetLoss := data.packetLoss[len(data.packetLoss)-1]
```
a-palchikov, line 243: Should take into account that `packetLoss` maybe empty.

monitoring/nethealth.go
```diff
@@ -142,18 +144,19 @@ func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter)
...
-	if len(pods.Items) == 0 {
-		return addr, trace.NotFound("unable to find local nethealth pod")
+	for _, pod := range pods.Items {
+		if pod.Status.HostIP == c.AdvertiseIP {
+			return fmt.Sprintf("http://%s:%d", pod.Status.PodIP, c.NethealthPort), nil
+		}
 	}
-	return fmt.Sprintf("http://%s:%d", pods.Items[0].Status.PodIP, c.NethealthPort), nil
+
+	return addr, trace.NotFound("unable to find local nethealth pod")
```
r0mant, line 159: Nit: I would include the IP we were looking for (`c.AdvertiseIP`) in the message.

monitoring/nethealth.go
```diff
@@ -260,6 +271,18 @@ func (c *nethealthChecker) isHealthy(peer string) (healthy bool, err error) {
 	return false, nil
 }
 
+// nethealthFailureProbe constructs a probe that represents failed nethealth check
+// against the specified peer.
+func nethealthFailureProbe(name, peer string, packetLoss float64) *pb.Probe {
+	return &pb.Probe{
+		Checker: name,
+		Detail: fmt.Sprintf("overlay packet loss for peer %s is higher than the allowed threshold of %.2f%%: %.2f%%",
```
r0mant, line 279: "peer" is too low-level as this is a user-facing message. Let's say "node".

r0mant (approved):
lgtm with a couple of comments + Dima's comments.

PR #181, opened 2020-04-01
[5.5.x] Fix sequence overflow bug

bernardjkim:
Marshalling an icmp message into an echo request converts the sequence number into uint16. This causes nethealth to think that the received echo responses are out of order once the echoCounter reaches 65535.

- Convert sequence number and echoCounter to uint16 when comparing.
- Log only echo counter for consistency even though it may not reflect
the actual sequence number used in the icmp echo request.

### DIFF
--- a/lib/nethealth/nethealth.go
+++ b/lib/nethealth/nethealth.go
@@ -487,7 +487,7 @@ func (s *Server) processAck(e messageWrapper) error {
 		return nil
 	default:
 		//unexpected / unknown
-		return trace.BadParameter("received unexpected icmp message type").(trace.Error).AddField("type", e.message.Type)
+		return trace.BadParameter("received unexpected icmp message type").AddField("type", e.message.Type)
 	}
 
 	switch pkt := e.message.Body.(type) {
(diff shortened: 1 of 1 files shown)


lib/nethealth/nethealth.go
```diff
@@ -496,8 +496,9 @@ func (s *Server) processAck(e messageWrapper) error {
 		if err != nil {
 			return trace.Wrap(err)
 		}
-		if pkt.Seq != peer.echoCounter {
-			return trace.BadParameter("Response sequence doesn't match latest request.")
+		if uint16(pkt.Seq) != uint16(peer.echoCounter) {
```
a-palchikov, line 499: nit: Could you also remove the type assertion from the `default` case above?
```go
default:
	// unexpected / unknown
	return trace.BadParameter("received unexpected icmp message type").AddField("type", e.message.Type)  // <- HERE
}
```

a-palchikov (approved):

lib/nethealth/nethealth.go
```diff
@@ -496,8 +496,9 @@ func (s *Server) processAck(e messageWrapper) error {
...
+			return trace.BadParameter("response sequence doesn't match latest request. Expected %d, received %d",
```
knisbet, line 500: Probably should add the variables as fields to the error instead of formatting as a string.

Something along the lines of:
```
return trace.BadParameter("response sequence doesn't match latest request.").AddFields(map[string]interface{}{
    "expected": uint16(peer.echoCounter),
    "received": uint16(peer.echoRequest),
}
```

knisbet (approved):

lib/nethealth/nethealth.go
```diff
@@ -508,7 +509,7 @@ func (s *Server) processAck(e messageWrapper) error {
 		s.WithFields(logrus.Fields{
 			"peer_name": peer.name,
 			"peer_addr": peer.addr,
-			"id":        pkt.Seq,
+			"id":        peer.echoCounter,
```
r0mant, line 512: Should this be converted to uint16 as well? Otherwise what's in the logs won't match what's actually being compared. We can also include the overall "real" counter e.g.

```go
  "counter": peer.echoCounter,
  "seq": uint16(peer.echoCounter),
```

r0mant (approved):

PR #182, opened 2020-04-01
[5.5.x] Need to unlock if client connection exists

bernardjkim:
- While updating `getAgentClient` to avoid locking I/O, I forgot to unlock clients in the case that the client connection exists...

- Log not found error.

- Return not found error if nethealth pod IP has not yet been assigned.

TODO: I think I may have rushed and merged previous PRs too early. Will do some testing and make sure I didn't introduce any other bugs...

### DIFF
--- a/monitoring/timedrift.go
+++ b/monitoring/timedrift.go
@@ -292,6 +292,7 @@ func (c *timeDriftChecker) shouldCheckNode(node serf.Member) bool {
 func (c *timeDriftChecker) getAgentClient(ctx context.Context, node serf.Member) (client.Client, error) {
 	c.mu.Lock()
 	if conn, exists := c.clients[node.Addr.String()]; exists {
+		c.mu.Unlock()
 		return conn, nil
 	}
 	c.mu.Unlock()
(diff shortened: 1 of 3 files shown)


monitoring/timedrift.go
```diff
@@ -292,6 +292,7 @@ func (c *timeDriftChecker) shouldCheckNode(node serf.Member) bool {
 func (c *timeDriftChecker) getAgentClient(ctx context.Context, node serf.Member) (client.Client, error) {
 	c.mu.Lock()
 	if conn, exists := c.clients[node.Addr.String()]; exists {
```
r0mant, line 294: There may still be a race condition. Consider this scenario:

* First caller comes here and sees there's no client yet. So it proceeds below and starts dialing.
* While it dials, the second caller comes here as well and also sees there's no client (because 1st one is still dialing). So it also proceeds below and starts dialing.
* Both establish the connection and the "last one wins" and the other one leaks.

I think to fix that we can check for the client presence again in the second Lock/Unlock section. And if the client is already there, close the one we've just created and return the one from the map. Does it make sense?

r0mant (approved):

PR #184, opened 2020-04-14
Log nethealth not found errors at debug level

bernardjkim:
Change log level of nethealth not found errors to DEBUG. This will prevent the logs from being flooded with not found errors if the monitoring capabilities are disabled on a cluster.

### DIFF
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -121,7 +121,7 @@ func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter) 
 func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter) error {
 	addr, err := c.getNethealthAddr()
 	if trace.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


monitoring/nethealth.go
```diff
@@ -121,7 +121,7 @@ func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter)
 func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter) error {
 	addr, err := c.getNethealthAddr()
 	if trace.IsNotFound(err) {
-		log.WithError(err).Warn("Nethealth pod was not found.")
+		log.WithError(err).Debug("Nethealth pod was not found.")
```
r0mant, line 124: I would honestly get rid of `.WithError(err)` as well - the stack trace is very verbose, so even at debug level not much benefit in seeing the whole stack trace for the not found error here.

r0mant (approved):

PR #186, opened 2020-04-14
System pods checker

bernardjkim:
###  Description 
This PR adds a system pods checker. This checker verifies that all pods with the `gravitational.io/critical-pod` label is healthy.

### Implementation
The checker queries k8s for all pods with `gravitational.io/critical-pod` label. The list of pods is further filtered to only contain pods that are running locally on that node.

The checker then verifies that the pods are healthy.

We start by checking the pod phase.
- If the pod has `Succeeded`, this indicates a completed job, so the checker considers this healthy.
- If the pod has `Failed`, the checker reports the pod as being unhealthy. 
- If the pod is `Unknown`, log this situation and continue. Will not report degraded probe.
- If the pod is `Running`, the checker then verifies that all containers are `Running` or in an OK state. 
- If the pod is `Pending`, the checker then checks to see if the pod is `Initialized`.
  - If the pod is `Initialized` then the checker verifies that all containers are `Running` or in an OK state.
  - If the pod is not `Initialized` then the checker verifies that all initContainers are `Running` or in an OK state. 

Container is unhealthy if:
- `Terminated:Error` - Container terminated with an error.
- `Waiting:CrashLoopBackOff` - Container is in a crash loop.
- `Waiting:ImagePullBackOff` - Container failed to pull image.
- `Waiting:ErrImagePull` - Container failed to pull image.

### Limitations
The current implementation is a bit conservative on what it determines as an unhealthy pod. Common error states, such as, `CrashLoopBackOff` and `ImagePullBackOff` will be reported as unhealthy, but other unhealthy states may go unnoticed. It won't take much work to add additional states to the list of healthy states, so we can add them in later.

The system-pods check will not be able to report on critical pods that have not been scheduled to a pod. Need to think about this scenario some more and will look to add this feature in the future.

### DIFF
--- a/lib/kubernetes/constants.go
+++ b/lib/kubernetes/constants.go
@@ -1,0 +1,20 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubernetes
+
+// AllNamespaces can be used to query pods in all namespaces.
+const AllNamespaces = ""
--- a/lib/kubernetes/constants.go
+++ b/lib/kubernetes/constants.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Kubernetes Authors.
+Copyright 2020 Gravitational, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/monitoring/system_pods.go
+++ b/monitoring/system_pods.go
@@ -135,25 +135,59 @@ func verifyPodStatus(status corev1.PodStatus) error {
 	switch status.Phase {
 	case corev1.PodPending, corev1.PodSucceeded:
 		return nil
+	case corev1.PodRunning:
+		return trace.Wrap(verifyConditions(status.Conditions))
 	case corev1.PodFailed:
 		return trace.BadParameter("pod has failed")
 	case corev1.PodUnknown:
 		log.Warn("Pod is in unknown state.")
 		return nil
+	default:
+		log.WithField("phase", status.Phase).Warn("Pod is in invalid phase.")
+		return trace.BadParameter("pod is in invalid phase")
 	}
+}
 
-	// https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions
-	// If the pod phase is `Running`, all conditions are expected to be true?
-	for _, condition := range status.Conditions {
-		if condition.Status == corev1.ConditionFalse {
-			return trace.BadParameter("%s condition is false; reason: %s", condition.Type, condition.Reason)
-		}
-		if condition.Status == corev1.ConditionUnknown {
-			log.Warnf("%s condition is in an unknown state.", condition.Type)
+// verifyConditions verifies all pod conditions.
+func verifyConditions(conditions []corev1.PodCondition) error {
+	for _, condition := range conditions {
+		if err := verifyCondition(condition); err != nil {
+			return trace.Wrap(err)
 		}
 	}
-
 	return nil
+}
+
+// verifyCondition verifies the provided pod condition.
+// The pod is expected to have `Initialized`, `ContainersReady`, and
+// `PodScheduled` set to true. The pod does not need to be `Ready`.
+func verifyCondition(condition corev1.PodCondition) error {
+	// https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions
+	switch condition.Type {
+	case corev1.PodScheduled, corev1.PodInitialized, corev1.ContainersReady:
+		return trace.Wrap(verifyConditionIsTrue(condition))
+	case corev1.PodReady:
+		return nil // `Running` pods are not always expected to be `Ready`. e.g. `gravity-site`.
+	default:
+		log.WithField("condition", condition.Type).Warnf("Received invalid pod condition.")
+		return trace.BadParameter("received invalid pod condition: %s", condition.Type)
+	}
+}
+
+// verifyConditionIsTrue verifies that the provided condition status is true.
+func verifyConditionIsTrue(condition corev1.PodCondition) error {
+	switch condition.Status {
+	case corev1.ConditionTrue:
+		return nil
+	case corev1.ConditionFalse:
+		return trace.BadParameter("%s condition is false; reason: %s", condition.Type, condition.Reason)
+	case corev1.ConditionUnknown:
+		log.Warnf("%s condition is in an unknown state.", condition.Type)
+		return nil
+	default:
+		log.WithField("status", condition.Status).Warnf("%s condition is in an invalid state", condition.Type)
+		return trace.BadParameter("%s condition is in an invalid state: %s", condition.Type, condition.Status)
+	}
 }
 
 const systemPodsCheckerID = "system-pods-checker"
--- a/monitoring/system_pods.go
+++ b/monitoring/system_pods.go
@@ -90,7 +90,7 @@ func (r *systemPodsChecker) Check(ctx context.Context, reporter health.Reporter)
 func (r *systemPodsChecker) check(ctx context.Context, reporter health.Reporter) error {
 	pods, err := r.getPods()
 	if trace.IsNotFound(err) {
-		log.WithError(err).Debug("Failed to get system pods.")
+		log.Debug("Failed to get system pods.")
 		return nil // system pods were not found, log and treat gracefully
 	}
 	if err != nil {
--- a/monitoring/system_pods.go
+++ b/monitoring/system_pods.go
@@ -82,7 +82,14 @@ func (r *systemPodsChecker) Check(ctx context.Context, reporter health.Reporter)
 	err := r.check(ctx, reporter)
 	if err != nil {
 		log.WithError(err).Warn("Failed to verify critical system pods")
-		reporter.Add(NewProbeFromErr(r.Name(), "failed to verify critical system pods", err))
+		// TODO: set probe to critical
+		reporter.Add(&pb.Probe{
+			Checker:  r.Name(),
+			Detail:   "failed to verify critical system pods",
+			Error:    trace.UserMessage(err),
+			Status:   pb.Probe_Failed,
+			Severity: pb.Probe_Warning,
+		})
 		return
 	}
 	if reporter.NumProbes() == 0 {
@@ -193,9 +200,10 @@ func verifyConditionIsTrue(condition corev1.PodCondition) error {
 // check for the pod specified by podName and namespace.
 func systemPodsFailureProbe(checkerName, podName, namespace string) *pb.Probe {
 	return &pb.Probe{
-		Checker: checkerName,
-		Detail:  fmt.Sprintf("pod %s running in namespace %s is in an invalid state", podName, namespace),
-		Status:  pb.Probe_Failed,
+		Checker:  checkerName,
+		Detail:   fmt.Sprintf("pod %s running in namespace %s is in an invalid state", podName, namespace),
+		Status:   pb.Probe_Failed,
+		Severity: pb.Probe_Warning, // TODO: set probe to critical
 	}
 }
 
--- a/monitoring/system_pods.go
+++ b/monitoring/system_pods.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/gravitational/satellite/agent/health"
 	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+	"github.com/gravitational/satellite/lib/kubernetes"
 	"github.com/gravitational/satellite/utils"
 
 	"github.com/gravitational/trace"
@@ -117,7 +118,7 @@ func (r *systemPodsChecker) getPods() ([]corev1.Pod, error) {
 		LabelSelector: systemPodsSelector.String(),
 		FieldSelector: fields.OneTermEqualSelector("spec.nodeName", r.NodeName).String(),
 	}
-	pods, err := r.Client.CoreV1().Pods("").List(opts)
+	pods, err := r.Client.CoreV1().Pods(kubernetes.AllNamespaces).List(opts)
 	if err != nil {
 		return nil, utils.ConvertError(err)
 	}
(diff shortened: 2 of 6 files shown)


r0mant:
@bernardjkim I think the design of this checker should be a little more flexible. I.e. I would design it so it accepts a list of filters for pods it should be checking for, and then have the caller (planet agent in our case) pass these filters. Checking for everything in kube-system is a bit too rigid - i.e. if something like logging application fails we probably shouldn't necessarily be degrading the cluster but if gravity-site or coredns fails, then yes. Same thing if we need to check something outside of kube-system namespace (we also have stuff in monitoring).

monitoring/system_pods.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/gravitational/satellite/agent/health"
+	"github.com/gravitational/satellite/utils"
+
+	"github.com/gravitational/trace"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SystemPodsConfig specifies configuration for a system pods checkers.
```
a-palchikov, line 32: checker~~s~~

monitoring/system_pods.go
```diff
@@ -0,0 +1,160 @@
...
+type SystemPodsConfig struct {
+	// AdvertiseIP specifies the advertised ip address of the host running this checker.
+	AdvertiseIP string
+	// KubeConfig specifies kubernetes access information.
```
a-palchikov, line 36: information -> configuration

monitoring/system_pods.go
```diff
@@ -0,0 +1,160 @@
...
+	*KubeConfig
+}
+
+// CheckAndSetDefaults validates that this configuration is correct and sets
+// value defaults where necessary.
+func (r *SystemPodsConfig) CheckAndSetDefaults() error {
```
a-palchikov, line 42: usually this is invoked in the constructor function and is not meant for the caller - I'd unexport it.

monitoring/system_pods.go
```diff
@@ -0,0 +1,160 @@
...
+
+	r.verifyPods(pods, reporter)
+	return nil
+}
+
+// getPods returns a list of the local pods that exist in the
+// systemPodsNamespace.
+func (r *systemPodsChecker) getPods() ([]corev1.Pod, error) {
+	opts := metav1.ListOptions{}
+	pods, err := r.Client.CoreV1().Pods(systemPodsNamespace).List(opts)
```
a-palchikov, line 108: I agree - the `systemPodsNamespace` needs to be configuration.

monitoring/system_pods.go
```diff
@@ -0,0 +1,160 @@
...
+	case corev1.PodUnknown:
+		log.Warn("Pod is in unknown state.")
+		return nil
+	}
+
+	// https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions
+	// If the pod phase is `Running`, all conditions are expected to be true?
+	for _, condition := range status.Conditions {
+		if condition.Status == corev1.ConditionFalse {
+			return trace.BadParameter("%s condition is false; reason: %s", condition.Type, condition.Reason)
```
a-palchikov, line 149: This will fail for gravity-site Pods where for each master node, there's an instance running (as part of a DaemonSet), but only the leader is `Ready`

bernardjkim: Ah, I see. I've made some changes in cda18c6. 

When verifying pod status we no longer verify the `Ready` condition, but still make sure that the pod the conditions `Initialized`, `ContainersReady`, `PodScheduled` are set to true.

bernardjkim:
How flexible should we make the pod filter?

Design A:

Allow planet to specify a list of queries that contain a namespace and a set of labels.
This approach allows us to be more precise with which pods to include in the checker, but will require more verbose configuration from planet.

```golang
// planet/.../checkers.go

queries := []monitoring.PodQuery{
	{
		Namespace: "kube-system",
		Labels: map[string]string{
			"app": "gravity-site",
		},
	},
	{
		Namespace: "monitoring",
		Labels: map[string]string{
			"app":       "monitoring",
			"component": "grafana",
		},
	},
        {
		Namespace: "monitoring",
		Labels: map[string]string{
			"app":       "alertmanager",
		},
	},	
}
[...]
```

Design B:

If you guys think we won't need to be that precise with the filter we could just have planet provide a set of namespaces and a set of label values. We can search each specified namespace for a predefined set of labels, (app, k8s-app, component, ...). This will make the configuration from planet a bit simpler.

```golang
// planet/.../checkers.go

systemPodsChecker, err := monitoring.NewSystemPodsChecker(
	monitoring.SystemPodsConfig{
		AdvertiseIP: config.AdvertiseIP,
		Namespaces: []string{
                        "kube-system",
                        "monitoring",
                },
		Pods:  []string{
                        "gravity-site",
                        "grafana",
                        "alertmanager",
                },
		KubeConfig:  &kubeConfig,
	},
)
```

a-palchikov:
What about flipping the sides - we could use a new label that will unify all relevant resources (pods). Then we'll need no filter at all, and all new system resources will be automatically accounted for w/o changes to planet checker code?

bernardjkim:
That probably makes the most sense. 02b0078

We can search through all namespaces and find all pods with a `system-pod` label.

monitoring/system_pods.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if reporter.NumProbes() == 0 {
+		reporter.Add(NewSuccessProbe(r.Name()))
+	}
+}
+
+func (r *systemPodsChecker) check(ctx context.Context, reporter health.Reporter) error {
+	pods, err := r.getPods()
+	if trace.IsNotFound(err) {
+		log.Debug("Failed to get system pods.")
```
a-palchikov, line 93: The message sort of contradicts the condition - no system pods found?

monitoring/system_pods.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// getPods returns a list of the local pods that have the system pod label.
+func (r *systemPodsChecker) getPods() ([]corev1.Pod, error) {
+	opts := metav1.ListOptions{
+		LabelSelector: systemPodsSelector.String(),
+	}
+	pods, err := r.Client.CoreV1().Pods("").List(opts)
+	if err != nil {
+		return nil, utils.ConvertError(err) // this will convert error to a proper trace error, e.g. trace.NotFound
```
a-palchikov, line 111: nit: the comment is redundant really.

monitoring/system_pods.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	return localPods, nil
+}
+
+// verifyPods verifies the pods are in a valid state. Reports a failed probe for
+// any pods that are in an invalid state.
+func (r *systemPodsChecker) verifyPods(pods []corev1.Pod, reporter health.Reporter) {
+	for _, pod := range pods {
+		if err := verifyPodStatus(pod.Status); err != nil {
+			reporter.Add(NewProbeFromErr(r.Name(), fmt.Sprintf("%s is in an invalid state", pod.Name), err))
```
a-palchikov, line 128: nit: include namespace with the message?

a-palchikov (approved):

monitoring/system_pods.go
```diff
@@ -0,0 +1,218 @@
...
+	for _, condition := range conditions {
+		if err := verifyCondition(condition); err != nil {
+			return trace.Wrap(err)
+		}
+	}
+	return nil
+}
+
+// verifyCondition verifies the provided pod condition.
+// The pod is expected to have `Initialized` and`PodScheduled` set to true.
```
r0mant, line 168: Nit: No space between "and" and PodScheduled.

monitoring/system_pods.go
```diff
@@ -0,0 +1,218 @@
...
+		return trace.BadParameter("%s condition is in an invalid state: %s", condition.Type, condition.Status)
+	}
+}
+
+// systemPodsFailureProbe constructs a probe that represents a failed system pods
+// check for the pod specified by podName and namespace.
+func systemPodsFailureProbe(checkerName, podName, namespace string) *pb.Probe {
+	return &pb.Probe{
+		Checker:  checkerName,
+		Detail:   fmt.Sprintf("pod %s running in namespace %s is in an invalid state", podName, namespace),
```
r0mant, line 204: I think this is missing the actual pod state (e.g. CrashLoopBackOff, Error, etc.). Can this be added? I'd also make the message a bit shorter e.g.

```
fmt.Sprintf("pod %v/%v is not Running: %v", namespace, name, <actual state>)
```

monitoring/system_pods.go
```diff
@@ -0,0 +1,218 @@
...
+			reporter.Add(systemPodsFailureProbe(r.Name(), pod.Name, pod.Namespace))
+		}
+	}
+}
+
+// verifyPodStatus verifies the status phase and conditions.
+func verifyPodStatus(status corev1.PodStatus) error {
+	// https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
+	switch status.Phase {
+	case corev1.PodPending, corev1.PodSucceeded:
```
r0mant, line 142: Regarding Pending status - do you know if there's a way to distinguish b/w when the pod is just pending to be scheduled versus when it's forever pending due to, for example, being unable to attach a persistent volume? While the former is ok, the latter case might be worth highlighting. Not sure there's a way to tell b/w the two though, I would just try it out and see.

bernardjkim:
@r0mant @a-palchikov 
I've revised the pod verification implementation 9ac343e. After running some tests, I realized the pod `status` isn't that straight forward. For example, `Running` pods should not always be considered healthy. A pod can be `Running` but have a container stuck in a `Waiting` state because of a `CrashLoopBackOff`. 

I've added additional steps to examine the state of the containers and initContainers in a pod. This will allow the checker more precision when determining if a pod is healthy/unhealthy.

I wasn't able to find much documentation on the container reasons (CrashLoopBackOff, ContainerCreating, etc), so I'm not exactly sure which ones are indicative of a unhealthy container/pod. I've added a few test cases of some of the different pod states I've observed while testing. Let me know if they look correct and if there are other common pod states that should be added to the unit tests. e8153a7

monitoring/system_pods.go
bernardjkim: So I'm not sure this checker can even verify either case. With the current implementation, if a pod is not yet scheduled to a node, the pod will not be verified. It will be filtered when we are querying local pods.

If we wanted to check pods that are unscheduled, we would need to modify the `getNodes` function or separately query pods that are not local to the node and verify them. All nodes would have to verify the same pods and we'd end up with duplicate probes on all nodes. 

If this is worth including, we could just have generic message saying there are some critical probes that have not been scheduled. Something like:

```
Cluster nodes:
    Masters:
        * node-1 / 172.28.128.101 / node
            Status:             healthy
            [!]                 One more more critical system pods have not been scheduled
            Remote access:      online
        * node-3 / 172.28.128.103 / node
            Status:             healthy
            [!]                 One more more critical system pods have not been scheduled
            Remote access:      online
        * node-2 / 172.28.128.102 / node
            Status:             healthy
            [!]                 One more more critical system pods have not been scheduled
            Remote access:      online
```

Also not sure if we would be able to distinguish the two scenarios. I don't think there is any field that changes if the pod has been continually failing to be scheduled?
```
Name:           task-pv-pod-7d898d594f-dq5bz
Namespace:      default
Priority:       0
Node:           <none>
Labels:         app=pv
                gravitational.io/critical-pod=true
                pod-template-hash=7d898d594f
Annotations:    kubernetes.io/psp: coredns
                seccomp.security.alpha.kubernetes.io/pod: docker/default
Status:         Pending
IP:
IPs:            <none>
Controlled By:  ReplicaSet/task-pv-pod-7d898d594f
Containers:
  task-pv-container:
    Image:        nginx
    Port:         80/TCP
    Host Port:    0/TCP
    Environment:  <none>
    Mounts:
      /usr/share/nginx/html from task-pv-storage (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-kpn6m (ro)
Conditions:
  Type           Status
  PodScheduled   False
Volumes:
  task-pv-storage:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  task-pv-claim
    ReadOnly:   false
  default-token-kpn6m:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-kpn6m
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  <none>
Tolerations:     node.kubernetes.io/not-ready:NoExecute for 300s
                 node.kubernetes.io/unreachable:NoExecute for 300s
Events:
  Type     Reason            Age        From               Message
  ----     ------            ----       ----               -------
  Warning  FailedScheduling  <unknown>  default-scheduler  persistentvolumeclaim "task-pv-claim" not found
  Warning  FailedScheduling  <unknown>  default-scheduler  persistentvolumeclaim "task-pv-claim" not found
```

a-palchikov:
@bernardjkim Yes, the documentation about Pod/Container statuses is sparse at best.
Before they're documented properly we can study the kubelet's [bits](https://github.com/kubernetes/kubernetes/blob/v1.17.5/pkg/kubelet/kubelet_pods.go#L1460) that deal with them.

monitoring/system_pods.go
```diff
@@ -0,0 +1,213 @@
...
+	pods, err := r.Client.CoreV1().Pods(kubernetes.AllNamespaces).List(opts)
+	if err != nil {
+		return nil, utils.ConvertError(err)
+	}
+
+	return pods.Items, nil
+}
+
+// verifyPods verifies the pods are in a valid state. Reports a failed probe for
+// any pods that are in an invalid state.
```
a-palchikov, line 130: not super clear whether it's a single probe or for each failed pod.

monitoring/system_pods.go
```diff
@@ -0,0 +1,213 @@
...
+			if isInitialized(pod.Status.Conditions) {
+				err = verifyContainers(pod.Status.ContainerStatuses)
+			} else {
+				err = verifyContainers(pod.Status.InitContainerStatuses)
+			}
+			if err != nil {
+				reporter.Add(systemPodsFailureProbe(r.Name(), pod.Namespace, pod.Name, err))
+			}
+		case corev1.PodFailed:
+			err := trace.BadParameter("pod Failed: %v", pod.Status.Reason)
```
a-palchikov, line 151: Failed->failed?

monitoring/system_pods.go
```diff
@@ -0,0 +1,213 @@
...
+		if status.State.Waiting != nil {
+			return trace.BadParameter("%v Waiting: %v", status.Name, status.State.Waiting.Reason)
+		}
+	}
+	return nil
+}
+
+// verifyInitialization verifies that the pod has been Inititalized or
+// InitContainers are still Running.
+func verifyInitialization(pod corev1.Pod) error {
```
a-palchikov, line 183: This seems to be unused.

a-palchikov (approved):

lib/kubernetes/constants.go
```diff
@@ -0,0 +1,20 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
r0mant, line 2: Was this file copied from K8s source code?

r0mant (approved):

lib/kubernetes/constants.go
bernardjkim: Oh, oops. Yeah, looks like I copied the header from the kubernetes src.

PR #191, opened 2020-04-17
Filter nethealth data

bernardjkim:
## Description

This PR addresses issue with cluster degrading during rolling update due to nethealth check failure https://github.com/gravitational/gravity/issues/1403.

**Changes to nethealth checker**

The nethealth-checker now filters incoming nethealth data through `filterNetData` which removes data for nodes that are no longer members of the cluster. Ideally, nethealth should handle removing metrics for these nodes, but having this filter here is probably a good idea anyways.

**Changes to nethealth application**

The nethealth application now uses the nethealth pod's host IP when recording metrics. Previously, the node name would be assigned to `node_name` and `peer_name` label. We can't always rely on node name being equal to the host IP, so filtering by serf IP would be unreliable. If we store the host IP in the relevant labels, we can then reliably filter metrics using serf member IPs.

Nethealth now removes metrics for peers that have left the cluster. This change should fix the underlying problem.

## Type of change

* Bug fix (non-breaking change which fixes an issue)
* Regression fix (non-breaking change which fixes a regression)

## Linked tickets and other PRs

* Refs https://github.com/gravitational/gravity/issues/1403

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

**Before changes**

- Setup multi-node cluster.
- Degrade nethealth checker using iptables to drop traffic from one node to another
```
[vagrant@node-1 ~]$ sudo iptables -A INPUT -p udp -s 172.28.128.102 -j DROP
```

- Once nodes are in a degraded state due to nethealth checker remove one of the degraded nodes.
```
[vagrant@node-1 ~]$ sudo gravity remove node-2 --force
```

- After the node is removed, the nethealth check is stuck in this state.
```
[vagrant@node-1 ~]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         degraded
[...]
Cluster nodes:
    Masters:
            Status:     degraded
            [×]         overlay packet loss for node 172.28.128.102 is higher than the allowed threshold of 20.00%: 100.00% ()
        * node-3 (172.28.128.103, node)
            Status:     healthy
```

**After changes**
If we follow the same procedure:

- Drop udp from node-2 to node-1 and wait for nethealth check to fail.
```
[vagrant@node-1 installer]$ sudo iptables -A INPUT -p udp -s 172.28.128.102 -j DROP
[vagrant@node-1 installer]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         degraded
[...]
Cluster nodes:
    Masters:
        * node-2 (172.28.128.102, node)
            Status:     degraded
            [×]         overlay packet loss for node 172.28.128.101 is higher than the allowed threshold of 20.00%: 100.00% ()
        * node-3 (172.28.128.103, node)
            Status:     healthy
            Status:     degraded
            [×]         overlay packet loss for node 172.28.128.102 is higher than the allowed threshold of 20.00%: 100.00% ()
```

- Now remove node-2 from the cluster.
```
[vagrant@node-1 installer]$ sudo gravity remove node-2 --force
[vagrant@node-1 installer]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         degraded
[...]
Cluster nodes:
    Masters:
        * node-3 (172.28.128.103, node)
            Status:     healthy
            Status:     degraded
            [×]         overlay packet loss for node 172.28.128.102 is higher than the allowed threshold of 20.00%: 100.00% ()
```

- Wait for next status update cycle and nethealth check should now pass.
```
[vagrant@node-1 installer]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         active
[...]
Cluster nodes:
    Masters:
        * node-3 (172.28.128.103, node)
            Status:     healthy
            Status:     healthy
```

If we look at the Prometheus metrics exposed by nethealth, we can see that metrics for node-2 were removed when node-2 left the cluster.

```
[vagrant@node-1 installer]$ curl 10.244.52.9:9801/metrics
[...]
# HELP nethealth_echo_request_total The number of echo requests that have been sent
# TYPE nethealth_echo_request_total counter
nethealth_echo_request_total{node_name="172.28.128.101",peer_name="172.28.128.103"} 807
# HELP nethealth_echo_timeout_total The number of echo requests that have timed out
# TYPE nethealth_echo_timeout_total counter
nethealth_echo_timeout_total{node_name="172.28.128.101",peer_name="172.28.128.103"} 0
[...]
```

### DIFF
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/gravitational/roundtrip"
 	"github.com/gravitational/trace"
+	serf "github.com/hashicorp/serf/client"
 	"github.com/mailgun/holster"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
@@ -48,6 +49,8 @@ type NethealthConfig struct {
 	NethealthPort int
 	// NetStatsInterval specifies the duration to store net stats.
 	NetStatsInterval time.Duration
+	// SerfClient is the client to the local serf agent.
+	SerfClient agent.SerfClient
 	// KubeConfig specifies kubernetes access information.
 	*KubeConfig
 }
@@ -59,9 +62,13 @@ func (c *NethealthConfig) CheckAndSetDefaults() error {
 	if c.AdvertiseIP == "" {
 		errors = append(errors, trace.BadParameter("host advertise ip must be provided"))
 	}
+	if c.SerfClient == nil {
+		errors = append(errors, trace.BadParameter("local serf client can't be empty"))
+	}
 	if c.KubeConfig == nil {
 		errors = append(errors, trace.BadParameter("kubernetes access config must be provided"))
 	}
+
 	if c.NethealthPort == 0 {
 		c.NethealthPort = defaultNethealthPort
 	}
@@ -78,7 +85,7 @@ type nethealthChecker struct {
 	NethealthConfig
 	// Mutex locks access to peerStats
 	sync.Mutex
-	// peerStats maps a peer to its recorded nethealth stats.
+	// peerStats maps a peer IP to its recorded nethealth stats.
 	peerStats *netStats
 }
 
@@ -138,6 +145,12 @@ func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter) 
 		log.WithError(err).
 			WithField("nethealth-metrics", string(resp)).
 			Error("Received incomplete set of metrics. Could be due to a bug in nethealth or a change in labels.")
+		return nil
+	}
+
+	netData, err = c.filterNetData(netData)
+	if err != nil {
+		log.WithError(err).Error("Failed to filter nethealth data.")
 		return nil
 	}
 
@@ -394,6 +407,30 @@ func getPeerName(labels []*dto.LabelPair) (peer string, err error) {
 		}
 	}
 	return "", trace.NotFound("unable to find %s label", peerLabel)
+}
+
+// filterNetData filters the networkData. Nethealth may retain metrics for nodes
+// that have previously left the cluster. This data is now irrelevant and should
+// not be further processed.
+func (c *nethealthChecker) filterNetData(netData map[string]networkData) (filtered map[string]networkData, err error) {
+	members, err := c.SerfClient.Members()
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return filterBySerf(netData, members)
+}
+
+// filterBySerf removes netData for nodes that are not part of the serf cluster.
+func filterBySerf(netData map[string]networkData, members []serf.Member) (filtered map[string]networkData, err error) {
+	filtered = make(map[string]networkData)
+	for _, member := range members {
+		if data, exists := netData[member.Addr.String()]; exists {
+			filtered[member.Addr.String()] = data
+			continue
+		}
+		log.Warnf("Missing nethealth data for node %s", member.Addr.String())
+	}
+	return filtered, nil
 }
 
 // netStats holds nethealth data for a peer.
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -68,7 +68,6 @@ func (c *NethealthConfig) CheckAndSetDefaults() error {
 	if c.KubeConfig == nil {
 		errors = append(errors, trace.BadParameter("kubernetes access config must be provided"))
 	}
-
 	if c.NethealthPort == 0 {
 		c.NethealthPort = defaultNethealthPort
 	}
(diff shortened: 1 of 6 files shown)


monitoring/nethealth.go
```diff
@@ -59,9 +62,13 @@ func (c *NethealthConfig) CheckAndSetDefaults() error {
 	if c.AdvertiseIP == "" {
 		errors = append(errors, trace.BadParameter("host advertise ip must be provided"))
 	}
+	if c.SerfClient == nil {
+		errors = append(errors, trace.BadParameter("local serf client can't be empty"))
+	}
 	if c.KubeConfig == nil {
 		errors = append(errors, trace.BadParameter("kubernetes access config must be provided"))
 	}
+
```
r0mant, line 71: Nit: Bit inconsistent to have an empty line here compared to the rest of this method.

monitoring/nethealth.go
```diff
@@ -396,6 +409,30 @@ func getPeerName(labels []*dto.LabelPair) (peer string, err error) {
...
+		return nil, trace.Wrap(err)
+	}
+	return filterBySerf(netData, members)
+}
+
+// filterBySerf removes netData for nodes that are not part of the serf cluster.
+func filterBySerf(netData map[string]networkData, members []serf.Member) (filtered map[string]networkData, err error) {
+	filtered = make(map[string]networkData)
+	for _, member := range members {
+		if data, exists := netData[member.Addr.String()]; exists {
```
r0mant, line 427: We should be filtering out not just those that don't exist anymore. But those that are in left/failed state as well (or just "!= alive" for simplicity). If you look at the bug report, you will see that the member is actually a part of "serf members" but in a "left" state.

bernardjkim: Ah, you're right. Added check for serf member status. 0d27f7c

bernardjkim:
Nethealth previously recorded k8s node names for the node_name and peer_name metrics labels. We can't rely on serf member IP and node name to be consistent. 
 
45a3752 modifies nethealth to record host IP into the relevant labels so that the nethealth-checker filter the incoming metrics using the serf IP addresses.

After thinking about it some more, we could instead modify the nethealth-checker to query k8s nodes, and use the list of nodes to filter the incoming metrics. This wouldn't require changes to the nethealth application and leave less room for error?

a-palchikov:
@bernardjkim I'm in favor of your last proposal for ditching serf to query members and instead query nodes from kubernetes.

bernardjkim:
Ah, looks like worker nodes do not have permission to query k8s nodes.
I guess for now we'll stick with filter by serf.

bernardjkim:
Testing looks okay.

- Drop udp from node-2 to node-1 and wait for nethealth check to fail.
```
[vagrant@node-1 installer]$ sudo iptables -A INPUT -p udp -s 172.28.128.102 -j DROP
[vagrant@node-1 installer]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         degraded
[...]
Cluster nodes:
    Masters:
        * node-2 (172.28.128.102, node)
            Status:     degraded
            [×]         overlay packet loss for node 172.28.128.101 is higher than the allowed threshold of 20.00%: 100.00% ()
        * node-3 (172.28.128.103, node)
            Status:     healthy
            Status:     degraded
            [×]         overlay packet loss for node 172.28.128.102 is higher than the allowed threshold of 20.00%: 100.00% ()
```

- Now remove node-2 from the cluster 
```
[vagrant@node-1 installer]$ sudo gravity remove node-2 --force
[vagrant@node-1 installer]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         degraded
[...]
Cluster nodes:
    Masters:
        * node-3 (172.28.128.103, node)
            Status:     healthy
            Status:     degraded
            [×]         overlay packet loss for node 172.28.128.102 is higher than the allowed threshold of 20.00%: 100.00% ()
```

- Wait for next status update cycle and nethealth check should now pass.
```
[vagrant@node-1 installer]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         active
[...]
Cluster nodes:
    Masters:
        * node-3 (172.28.128.103, node)
            Status:     healthy
            Status:     healthy
```

The Prometheus metrics for node-2 were removed when node-2 left the cluster. If for some reason the metrics were not removed, it will at least have been filtered in the nethealth check.

monitoring/nethealth.go
```diff
@@ -396,6 +408,35 @@ func getPeerName(labels []*dto.LabelPair) (peer string, err error) {
...
+// filterBySerf removes netData for nodes that are not part of the serf cluster.
+func filterBySerf(netData map[string]networkData, members []serf.Member) (filtered map[string]networkData, err error) {
+	filtered = make(map[string]networkData)
+	for _, member := range members {
+		data, exists := netData[member.Addr.String()]
+		if !exists {
+			log.Warnf("Missing nethealth data for node %s", member.Addr.String())
+			continue
+		}
+		if agent.MemberStatus(member.Status) != agent.MemberAlive {
```
r0mant, line 431: So how does this work now if you block udp traffic? Presumably the serf member will become "failed" and the checker won't show anything? Maybe we should be skipping over "left" members only here?

r0mant (commented):
@a-palchikov @knisbet Could you guys take a look too?

monitoring/nethealth.go
bernardjkim: Looks like serf doesn't communicate through the overlay network, so they won't become `failed` in this situation. But yup, if tcp is blocked as well this would prevent the check from validating.

bernardjkim:
Ah, sorry for going back and forth on this one. Can we revisit proposal to query k8s nodes in the checker instead of using serf? I'll make a separate PR with implementation to filter by k8s nodes. The nethealth checker is already querying k8s pods, it should have permission to query the nodes.

r0mant:
> The nethealth checker is already querying k8s pods, it should have permission to query the nodes even if its running on a worker node.

@bernardjkim I'm not sure about this. Nodes are a cluster-wide resource and regular nodes do not have permissions to query them (to my recollection).

bernardjkim:
@r0mant From initial testing, it seems like worker nodes are successfully able to query nodes. I'll do some do some proper testing to verify and get back to you.

PR #193, opened 2020-04-28
Prevent nethealth check from degrading nodes

bernardjkim:
The nethealth check has been causing a lot of problems. I think it is probably best to prevent nethealth check from degrading nodes for now until we fix some issue with the checker. With this change, nethealth check will no longer report failed probes, instead all network issues will just be logged.

### DIFF
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -110,7 +110,9 @@ func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter) 
 	err := c.check(ctx, reporter)
 	if err != nil {
 		log.WithError(err).Warn("Failed to verify nethealth")
-		reporter.Add(NewProbeFromErr(c.Name(), "failed to verify nethealth", err))
+
+		// Do not report failed probes for now until we fix issues with nethealth checker.
+		// reporter.Add(NewProbeFromErr(c.Name(), "failed to verify nethealth", err))
 		return
 	}
 	if reporter.NumProbes() == 0 {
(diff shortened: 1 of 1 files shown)


monitoring/nethealth.go
```diff
@@ -110,7 +110,9 @@ func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter)
 	err := c.check(ctx, reporter)
 	if err != nil {
 		log.WithError(err).Warn("Failed to verify nethealth")
-		reporter.Add(NewProbeFromErr(c.Name(), "failed to verify nethealth", err))
+
+		// Do not report failed probes for now until we fix issues with nethealth checker.
+		// reporter.Add(NewProbeFromErr(c.Name(), "failed to verify nethealth", err))
```
r0mant, line 115: Let's not leave commented code. Let's replace with a TODO item or something.

monitoring/nethealth.go
```diff
@@ -261,6 +264,34 @@ func (c *nethealthChecker) verifyNethealth(peers []string, reporter health.Repor
...
+		// Report last recorded packet loss percentage
+		data, err := c.peerStats.Get(peer)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		if len(data.packetLoss) == 0 {
+			continue
+		}
+		packetLoss := data.packetLoss[len(data.packetLoss)-1]
+		log.Errorf("Overlay packet loss for node %s is higher than the allowed threshold of %.2f%%: %.2f%%",
```
r0mant, line 289: I'd make this a warning.

bernardjkim:
Simpler to just disable checks in planet.

PR #195, opened 2020-05-04
[5.5.x] Warning severity no longer degrades nodes

bernardjkim:
### Description 
This PR modifies satellite to no longer degrade nodes for probes with a `Warning` severity. Implemented in `master` at https://github.com/gravitational/satellite/pull/112. 

### Purpose 
Enabling `Warning` probes can be used to help reduce the amount of risk when introducing new health checks to gravity. New health checks can be initially added as warnings. Once they have been seen to function as expected, the check can then be switched to critical.

### DIFF
--- a/agent/health/health.go
+++ b/agent/health/health.go
@@ -99,7 +99,7 @@ func (r Probes) GetFailed() []*pb.Probe {
 func (r Probes) Status() pb.NodeStatus_Type {
 	result := pb.NodeStatus_Running
 	for _, probe := range r {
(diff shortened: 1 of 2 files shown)


r0mant (approved):
Do you know if we have any existing "warning" probes? Or just the timedrift that you fixed here?

bernardjkim:
@r0mant 
Looks like `ping` check and `nodestatus` check are using warnings. They should still be degrading the node on failure. Will update.

a-palchikov (approved):

PR #198, opened 2020-05-06
Use distroless base image

bernardjkim:
### Description
This PR replaces the ubuntu 19.10 base image used when building the nethealth image with a distroless base image.

### Purpose
Security scans for the nethealth images built on top of ubuntu 19.10 are reporting some vulnerabilities. Moving to distroless can help reduce these vulnerabilities. Though it looks like quay does not currently support security scans for distroless images.

### DIFF
--- a/Dockerfile.nethealth
+++ b/Dockerfile.nethealth
@@ -17,7 +17,7 @@ ARG BUILD_VERSION
 ARG RIGGING_IMAGE
 
 FROM $BUILD_IMAGE as build
-RUN apt-get update && apt-get install -y libcap2-bin && apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN set -ex && apt-get update && apt-get install -y libcap2-bin && apt-get clean && rm -rf /var/lib/apt/lists/*
 ADD ./ /go/src/github.com/gravitational/satellite/
 RUN cd /go/src/github.com/gravitational/satellite/ && BUILD_VERSION=${BUILD_VERSION} go run mage.go build:nethealth
 
--- a/Dockerfile.nethealth
+++ b/Dockerfile.nethealth
@@ -17,16 +17,22 @@ ARG BUILD_VERSION
 ARG RIGGING_IMAGE
 
 FROM $BUILD_IMAGE as build
+RUN apt-get update && apt-get install -y libcap2-bin && apt-get clean && rm -rf /var/lib/apt/lists/*
 ADD ./ /go/src/github.com/gravitational/satellite/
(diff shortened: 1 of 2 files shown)


a-palchikov:
I don't know how and which security checks quay runs but it could be based off of some database of know vulnerabilities pinned to a specific distribution. With distroless, there is no distribution and if it does not check the binary (akin to an antivirus) - it won't be able to report anything.

a-palchikov (approved):

Dockerfile.nethealth
```diff
@@ -17,16 +17,22 @@ ARG BUILD_VERSION
...
+
 #
 # Build nethealth container
 #
 FROM ${BASE_IMAGE}
-RUN apt-get update && apt-get install -y libcap2-bin && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+COPY --from=build /sbin/getcap /sbin
+COPY --from=build /sbin/setcap /sbin
+COPY --from=build /lib/x86_64-linux-gnu/libcap.so.2 /lib
```
UnderGreen, line 32: That's copy actions are dangerous because base image and build image could be incompatible.

knisbet:
> I don't know how and which security checks quay runs but it could be based off of some database of known vulnerabilities pinned to a specific distribution. With distroless, there is no distribution and if it does not check the binary (akin to an antivirus) - it won't be able to report anything.

Yea, quay won't be able to scan distroless containers. Quay uses the clair scanner, which only supports certain packaging formats. However, distroless is the target here, because the surface area is extremely small. Depending on the one picked, it's basically just libc and then anything we add. This makes it extremely difficult to chain vulnerabilities or use the runtime for exploit, and the surface area is tiny, and exploiting just a libc vulnerability without anything else is difficult. 

The vulnerabilities being reported are likely not exploitable, but it does make people feel better to see a clean scan result.

Dockerfile.nethealth
knisbet: Yea, unfortunately capability bits don't seem to be copiable (I suspect this is a design choice, so that a copied binary doesn't accidentally maintain capabilities that it shouldn't have). So I don't think this is the end of the world, especially going into a distroless container.

knisbet (approved):

Dockerfile.nethealth
bernardjkim: So if its unsafe to copy libcap to our image, will nethealth be able to function the same without `cap_net_raw`?

r0mant (approved):

PR #199, opened 2020-05-07
Add RPC to update local timeline

bernardjkim:
### Description
This PR adds an additional RPC that can be used to record events to an agent's **local** timeline. 

### Purpose
This RPC can be used by `gravity-site` to record events into the timeline. Events for `add/join`, `remove/leave`, and `upgrade` might be worth including in the timeline.


a-palchikov (approved):

r0mant (commented):
Could you please remind what a local timeline vs cluster timeline is? :)

bernardjkim:
@r0mant In the current implementation of satellite we have two separate timelines. Events that were observed locally, such as changes to the node status and probe results are first recorded in the local timeline. These events are then pushed to all master nodes to be recorded in the cluster timeline. 
 
Only the cluster timeline is externally visible. This was our approach to avoid duplicate events in the timeline.

PR #204, opened 2020-05-19
Query members before running nethealth check

bernardjkim:
### Description
On a single node cluster, the logs are spammed with `Original Error: *trace.NotFoundError nethealth_echo_request_total metrics not found ...` messages. This is because the current impl expects nethealth nethealth request/timeout counters to be present in the incoming metrics. For a single node cluster, no counters should be initialized.

This single node scenario should be expected and handled without logging an error.

Requires backport to 5.5, 6.1, 7.0.

### Type of change
* Bug fix (non-breaking change which fixes an issue)

### TODOs
- [x] Manual testing
- [x] Address PR feedback

### Testing Done
**Verify nethealth check does not spam logs with not found errors**

_Before_
- Create a single node cluster.
- Examine `journalctl -u planet-agent` output.
- Logs are filled with not found errors:
```
May 16 12:37:35 node-1 /usr/bin/planet[929]: ERRO             Received incomplete set of metrics. Could be due to a bug in nethealth or a change in labels. error:[
                                             ERROR REPORT:
                                             Original Error: *trace.NotFoundError nethealth_echo_request_total metrics not found
                                             Stack Trace: [...]
```
_After_
- Logs are no longer filled with not found errors.

**Verify nethealth check still functions the same**
- After disable udp, nethealth should display warning.
```
[vagrant@node-1 installer]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         active
Application:            telekube, version 5.5.46-dev.2
Gravity version:        5.5.46-dev.2 (client) / 5.5.46-dev.2 (server)
[...]
Cluster nodes:
    Masters:
        * node-2 (172.28.128.102, node)
            Status:             healthy
            [!]                 overlay packet loss for node 172.28.128.101 is higher than the allowed threshold of 20%: 100% ()
            Remote access:      online
        * node-3 (172.28.128.103, node)
            Status:             healthy
            Remote access:      online
            Status:             healthy
            [!]                 overlay packet loss for node 172.28.128.102 is higher than the allowed threshold of 20%: 100% ()
            Remote access:      online
```

** Verfiy incoming data is still filtered when a node leaves the cluster**
- Remove node and wait for status to update. `node-1` should no longer show nethealth warning.
```
[vagrant@node-1 installer]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         active
Application:            telekube, version 5.5.46-dev.2
Gravity version:        5.5.46-dev.2 (client) / 5.5.46-dev.2 (server)
[...]
Cluster nodes:
    Masters:
        * node-3 (172.28.128.103, node)
            Status:             healthy
            Remote access:      online
            Status:             healthy
            Remote access:      online
```

### DIFF
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -142,6 +142,10 @@ func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter) 
 	}
 
 	netData, err := parseMetrics(resp)
+	if trace.IsNotFound(err) {
+		log.Debug("Nethealth counters are uninitialized.")
+		return nil
+	}
 	if err != nil {
 		log.WithError(err).
 			WithField("nethealth-metrics", string(resp)).
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -127,6 +127,15 @@ func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter) 
 }
 
 func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter) error {
+	peers, err := c.getPeers()
+	if err != nil {
+		log.Debug("Failed to discover nethealth peers.")
+		return nil
+	}
+	if len(peers) == 0 {
+		return nil
+	}
+
 	addr, err := c.getNethealthAddr()
 	if trace.IsNotFound(err) {
 		log.Debug("Nethealth pod was not found.")
@@ -142,10 +151,6 @@ func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter) 
 	}
 
 	netData, err := parseMetrics(resp)
-	if trace.IsNotFound(err) {
-		log.Debug("Nethealth counters are uninitialized.")
-		return nil
-	}
 	if err != nil {
 		log.WithError(err).
 			WithField("nethealth-metrics", string(resp)).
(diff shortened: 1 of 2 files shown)


r0mant (commented):
Instead of suppressing the error, why don't we just check how many nodes there currently are in the cluster and have the checker do nothing (return) if there's only one? That's basically how other multi-node checkers work.

monitoring/nethealth.go
```diff
@@ -127,6 +127,15 @@ func (c *nethealthChecker) Check(ctx context.Context, reporter health.Reporter)
 }
 
 func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter) error {
+	peers, err := c.getPeers()
+	if err != nil {
+		log.Debug("Failed to discover nethealth peers.")
```
a-palchikov, line 132: Even if you could just embed the error message - it would still be better than without:
```go
log.Debugf("Failed to discover nethealth peers: %v.", err)
```

monitoring/nethealth.go
```diff
@@ -142,31 +151,37 @@ func (c *nethealthChecker) check(ctx context.Context, reporter health.Reporter)
...
 	}
 
 	return c.verifyNethealth(updated, reporter)
 }
 
+// getPeers returns all nethealth peers as a list of strings.
+func (c *nethealthChecker) getPeers() (peers []string, err error) {
+	opts := metav1.ListOptions{
+		LabelSelector: nethealthLabelSelector.String(),
+		FieldSelector: fields.OneTermNotEqualSelector("metadata.name", c.NodeName).String(),
```
a-palchikov, line 173: I was confused at first - this query was used for nodes, not pods. It will not work for pods as far as I can tell - did you want `spec.nodeName` instead?

a-palchikov:
This should be backported to 5.5.x too.

monitoring/nethealth.go
bernardjkim: Ah, yup. This should `spec.nodeName`

a-palchikov (approved):

PR #218, opened 2020-06-02
Revert golang version

bernardjkim:
### Description
This PR updates the ping time format. The latency times will now be output using the largest time unit without a leading 0.

### DIFF
--- a/monitoring/ping.go
+++ b/monitoring/ping.go
@@ -193,22 +193,18 @@ func (c *pingChecker) checkNodesRTT(nodes []serf.Member, client agent.SerfClient
 			return trace.Wrap(err)
 		}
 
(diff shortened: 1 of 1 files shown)


build.go
```diff
@@ -29,11 +29,11 @@ import (
 
 var (
 	// buildContainer is a docker container used to build go binaries
-	buildContainer = "golang:1.14.3"
+	buildContainer = "golang:1.12.17"
```
knisbet, line 32: I think it might be legitimate to go through and bump golang on master, it's just we should be careful about the backports.

knisbet (approved):

build.go
bernardjkim: Oh, ok. I'll leave the changes in the master branch only then.

PR #219, opened 2020-06-02
[5.5.x] Revert golang version

bernardjkim:
### Description
This PR reverts golang version to 1.12.0. The ping time format is also updated. The latency times will now be output using the largest time unit without a leading 0.

### DIFF
--- a/build.go
+++ b/build.go
@@ -29,11 +29,11 @@ import (
 
 var (
 	// buildContainer is a docker container used to build go binaries
-	buildContainer = "golang:1.14.3"
+	buildContainer = "golang:1.12.0"
 
 	// golangciVersion is the version of golangci-lint to use for linting
 	// https://github.com/golangci/golangci-lint/releases
-	golangciVersion = "v1.27.0"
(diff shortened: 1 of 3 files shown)


build.go
```diff
@@ -29,11 +29,11 @@ import (
 
 var (
 	// buildContainer is a docker container used to build go binaries
-	buildContainer = "golang:1.14.3"
+	buildContainer = "golang:1.12.0"
 
 	// golangciVersion is the version of golangci-lint to use for linting
 	// https://github.com/golangci/golangci-lint/releases
-	golangciVersion = "v1.27.0"
+	golangciVersion = "v1.15.0"
```
knisbet, line 36: it's not clear to me if this is walking back a specific PR, but I think using a later golangciVersion isn't a problem here.

knisbet (approved):

build.go
bernardjkim: Reverting golang version changes from https://github.com/gravitational/satellite/pull/214. I think specific golangci versions are only compatible with specific golang versions. Will just revert both for now.

PR #224, opened 2020-06-05
Update disk check

bernardjkim:
### Description
This PR updates the disk check and addresses items 1-3 of https://github.com/gravitational/gravity/issues/1662. The disk/storage check will now report a warning or a critical probe at different thresholds. The default is set to 80% disk usage for a warning probe and 90% disk usage for a critical probe. 

### Linked tickets and PRs
* Updates https://github.com/gravitational/gravity/issues/1662

### Testing done
Testing done on version 5.5.47. `fallocate` is a quick way to alloate disk space for a file.

**Verify warning when disk usage > 80%**
```
[vagrant@node-1 gravity]$ df
Filesystem     1K-blocks     Used Available Use% Mounted on
[...]
/dev/sdb1       25670884 20356768   3987064  84% /var/lib/gravity

[vagrant@node-1 gravity]$ sudo gravity status
[...]
Cluster nodes:
    Masters:
            Status:             healthy
            [!]                 disk utilization on /var/lib/gravity exceeds 80 percent (4.1GB is available out of 26GB), see https://gravitational.com/telekube/docs/cluster/#garbage-collection
            Remote access:      online
[...]
```

**Verify critical when disk usage > 90%**
```
[vagrant@node-1 gravity]$ df
Filesystem     1K-blocks     Used Available Use% Mounted on
[...]
/dev/sdb1       25670884 22229512   2114320  92% /var/lib/gravity

[vagrant@node-1 gravity]$ sudo gravity status
[...]
Cluster nodes:
    Masters:
            Status:             degraded
            [×]                 disk utilization on /var/lib/gravity exceeds 90 percent (2.2GB is available out of 26GB), see https://gravitational.com/telekube/docs/cluster/#garbage-collection
            Remote access:      online
[...]
```

### DIFF
--- a/monitoring/storage.go
+++ b/monitoring/storage.go
@@ -19,6 +19,8 @@ package monitoring
 import (
 	"fmt"
 
+	"github.com/gravitational/trace"
+
 	humanize "github.com/dustin/go-humanize"
 )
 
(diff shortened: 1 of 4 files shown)


monitoring/storage.go
```diff
@@ -50,17 +67,26 @@ type HighWatermarkCheckerData struct {
...
 	return fmt.Sprintf("disk utilization on %s is below %v percent (%s is available out of %s)",
-		d.Path, d.HighWatermark, humanize.Bytes(d.AvailableBytes), humanize.Bytes(d.TotalBytes))
+		d.Path, d.WatermarkWarning, humanize.Bytes(d.AvailableBytes), humanize.Bytes(d.TotalBytes))
 }
 
 // DiskSpaceCheckerID is the checker that checks disk space utilization
 const DiskSpaceCheckerID = "disk-space"
+
+// DefaultCriticalWatermark is the default critical disk usage percentage threshold.
+const DefaultCriticalWatermark = 90
```
r0mant, line 92: I'd move this constants alongside the other one (the low watermark), I think it's somewhere in lib/constants.

monitoring/storage_linux.go
```diff
@@ -155,30 +159,48 @@ func (c *storageChecker) checkHighWatermark(ctx context.Context, reporter health
 		return trace.BadParameter("disk capacity at %v is 0", c.path)
 	}
 	checkerData := HighWatermarkCheckerData{
-		HighWatermark:  c.HighWatermark,
-		Path:           c.Path,
-		TotalBytes:     totalBytes,
-		AvailableBytes: availableBytes,
+		WatermarkCritical: c.HighWatermark,
+		WatermarkWarning:  c.HighWatermark - 10, // Set warning watermark 10% below the critical watermark
```
r0mant, line 163: I'd just have 2 separate constants TBH for soft/hard limit. Is there any particular reason you wanted to "tie" them one to another this way?

bernardjkim: We want the watermark to be configurable/overrideable in emergency situations. If the values are separately configurable, we might run into incorrect behavior if the critical watermark is set to a lower value than the warning watermark. Having them tied together will make sure we don't run into that situation.

r0mant: This scenario can be just checked during checker initialization, right? In Check(), as we usually do. And just return an error if that's the case.

monitoring/storage.go
bernardjkim: A watermark constant was not previously defined in satellite. The constant is defined in planet and passed to the storage checker.

monitoring/storage_linux.go
bernardjkim: Yeah, that works. I guess it'd be nicer to have separate control over the two threshold values.

r0mant (approved):

a-palchikov (approved):

a-palchikov:
This should be also taken into account: https://github.com/gravitational/gravity/issues/1748

PR #225, opened 2020-06-08
[5.5.x] Update disk check

bernardjkim:
### Description
This PR updates the disk check to report two separate warning and critical probes at different thresholds.

### Linked tickets and PRs
* Ports https://github.com/gravitational/satellite/pull/224

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -29,7 +29,7 @@ required = [
 
 [[constraint]]
   name = "github.com/gogo/protobuf"
-  version = "=v1.2.1"
+  version = "=v1.3.0"
 
 [[constraint]]
   name = "github.com/mattn/go-sqlite3"
--- a/monitoring/storage.go
+++ b/monitoring/storage.go
@@ -36,8 +36,10 @@ type StorageConfig struct {
 	Filesystems []string
 	// MinFreeBytes define minimum free volume capacity
 	MinFreeBytes uint64
-	// HighWatermark is the disk occupancy percentage that is considered degrading
-	HighWatermark uint
+	// WatermarkWarning is the disk occupancy percentage that will trigger a warning probe
+	WatermarkWarning uint
+	// WatermarkCritical is the disk occupancy percentage that will trigger a critical probe
+	WatermarkCritical uint
 }
 
 func (c *StorageConfig) CheckAndSetDefaults() error {
@@ -45,9 +47,23 @@ func (c *StorageConfig) CheckAndSetDefaults() error {
 	if c.Path == "" {
 		errors = append(errors, trace.BadParameter("volume path must be provided"))
 	}
-	if c.HighWatermark == 0 {
-		c.HighWatermark = DefaultCriticalWatermark
+
+	if c.WatermarkWarning > 100 {
+		errors = append(errors, trace.BadParameter("warning watermark must be between 0 and 100"))
 	}
+
+	if c.WatermarkCritical > 100 {
+		errors = append(errors, trace.BadParameter("critical watermark must be between 0 and 100"))
+	}
+
+	if c.WatermarkWarning == 0 {
+		c.WatermarkWarning = DefaultWarningWatermark
+	}
+
+	if c.WatermarkCritical == 0 {
+		c.WatermarkCritical = DefaultCriticalWatermark
+	}
+
 	return trace.NewAggregate(errors...)
 }
 
@@ -80,11 +96,14 @@ func (d HighWatermarkCheckerData) CriticalMessage() string {
 // SuccessMessage returns success watermark check message
 func (d HighWatermarkCheckerData) SuccessMessage() string {
 	return fmt.Sprintf("disk utilization on %s is below %v percent (%s is available out of %s)",
-		d.Path, d.WatermarkWarning, humanize.Bytes(d.AvailableBytes), humanize.Bytes(d.TotalBytes))
+		d.Path, d.WatermarkCritical, humanize.Bytes(d.AvailableBytes), humanize.Bytes(d.TotalBytes))
 }
 
 // DiskSpaceCheckerID is the checker that checks disk space utilization
 const DiskSpaceCheckerID = "disk-space"
+
+// DefaultWarningWatermark is the default warning disk usage percentage threshold.
+const DefaultWarningWatermark = 80
 
 // DefaultCriticalWatermark is the default critical disk usage percentage threshold.
 const DefaultCriticalWatermark = 90
--- a/monitoring/storage.go
+++ b/monitoring/storage.go
@@ -68,7 +68,7 @@ func (c *StorageConfig) CheckAndSetDefaults() error {
 	}
 
 	if c.LowWatermark > c.HighWatermark {
-		errors = append(errors, trace.BadParameter("low watermark (%v) cannot be higher than high watermark (%v)", c.LowWatermark, c.HighWatermark))
+		c.LowWatermark = c.HighWatermark
 	}
 
 	return trace.NewAggregate(errors...)
--- a/monitoring/storage.go
+++ b/monitoring/storage.go
@@ -42,6 +42,9 @@ type StorageConfig struct {
 	WatermarkCritical uint
 }
 
+// CheckAndSetDefaults validates this configuration object.
+// Config values that were not specified will be set to their default values if
+// available.
 func (c *StorageConfig) CheckAndSetDefaults() error {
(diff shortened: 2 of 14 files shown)


monitoring/storage.go
```diff
@@ -34,14 +36,43 @@ type StorageConfig struct {
...
+	if c.Path == "" {
+		errors = append(errors, trace.BadParameter("volume path must be provided"))
+	}
+
+	if c.WatermarkWarning > 100 {
+		errors = append(errors, trace.BadParameter("warning watermark must be between 0 and 100"))
+	}
+
+	if c.WatermarkCritical > 100 {
+		errors = append(errors, trace.BadParameter("critical watermark must be between 0 and 100"))
```
a-palchikov, line 56: Do you want to make sure Critical > Warning?

monitoring/storage.go
```diff
@@ -50,17 +81,29 @@ type HighWatermarkCheckerData struct {
 	AvailableBytes uint64 `json:"available_bytes"`
 }
 
-// FailureMessage returns failure watermark check message
-func (d HighWatermarkCheckerData) FailureMessage() string {
+// WarningMessage returns warning watermark check message
+func (d HighWatermarkCheckerData) WarningMessage() string {
+	return fmt.Sprintf("disk utilization on %s exceeds %v percent (%s is available out of %s), see https://gravitational.com/telekube/docs/cluster/#garbage-collection",
```
a-palchikov, line 86: I would probably mention that the cluster will degrade once it reaches %v (critical watermark).

monitoring/storage_linux.go
```diff
@@ -143,8 +147,13 @@ func (c *storageChecker) checkFsType(ctx context.Context, reporter health.Report
 	return nil
 }
 
+// checkHighWatermark checks the disk usage. A failed warning or critical probe
+// will be reported if the usage percentage is above the set thresholds.
+// If the WatermarkWarning percentage is higher than or equal to the
```
a-palchikov, line 152: Does it make sense to be able to set warning values exceeding the critical value?

monitoring/storage_linux.go
```diff
@@ -143,8 +147,13 @@ func (c *storageChecker) checkFsType(ctx context.Context, reporter health.Report
...
+// WatermarkCritical percentage, then the check will only ever report critical
+// probes.
 func (c *storageChecker) checkHighWatermark(ctx context.Context, reporter health.Reporter) error {
```
a-palchikov, line 155: I'd call this `checkDiskUsage` or something similar to emphasize the purpose.

monitoring/storage_linux_test.go
```diff
@@ -103,18 +103,20 @@ func (_ *StorageSuite) TestStorage(c *C) {
 
 	storageChecker{
 		StorageConfig: StorageConfig{
-			Path:          path.Join("/tmp", fmt.Sprintf("%d", time.Now().Unix())),
-			WillBeCreated: true,
-			HighWatermark: 40,
+			Path:              path.Join("/tmp", fmt.Sprintf("%d", time.Now().Unix())),
+			WillBeCreated:     true,
+			WatermarkCritical: 40,
```
a-palchikov, line 108: Not sure these values are useful. What about exercising happy path with warning < critical and disk usage reaching both warning, then criticial. Also warning >= critical which should yield an error, imo.

monitoring/storage.go
```diff
@@ -34,14 +36,43 @@ type StorageConfig struct {
...
 	MinFreeBytes uint64
-	// HighWatermark is the disk occupancy percentage that is considered degrading
-	HighWatermark uint
+	// WatermarkWarning is the disk occupancy percentage that will trigger a warning probe
+	WatermarkWarning uint
+	// WatermarkCritical is the disk occupancy percentage that will trigger a critical probe
+	WatermarkCritical uint
+}
+
+func (c *StorageConfig) CheckAndSetDefaults() error {
```
r0mant, line 45: Nit: Missing godoc.

monitoring/storage.go
r0mant, line 86: I think this link needs updating as there's no more telekube: https://gravitational.com/telekube/docs/cluster/#garbage-collection.

The current link is https://gravitational.com/gravity/docs/cluster/#garbage-collection. Same below.

r0mant (commented):
lgtm with Dima's comments

monitoring/storage.go
bernardjkim: The ui might be a little weird if we ensure that critical > warning. Example: if we set critical to 50 and leave warning unset, warning will default to 80. 

I guess what we can do is first set defaults. Then, if warning > critical just set warning == critical?

monitoring/storage_linux.go
```diff
@@ -147,15 +147,16 @@ func (c *storageChecker) checkFsType(ctx context.Context, reporter health.Report
...
-// probes.
-func (c *storageChecker) checkHighWatermark(ctx context.Context, reporter health.Reporter) error {
-	if c.WatermarkCritical == 0 {
+// checkDiskUsage checks the disk usage. A warning or critical probe will be
+// reported if the usage percentage is above the set thresholds.
+func (c *storageChecker) checkDiskUsage(ctx context.Context, reporter health.Reporter) error {
+	if c.HighWatermark == 0 {
 		return nil
 	}
+	if c.LowWatermark > c.HighWatermark {
```
a-palchikov, line 156: I think this is redundant as already being validated in `StorageConfig.CheckAndSetDefaults`?

monitoring/storage.go
```diff
@@ -36,43 +36,50 @@ type StorageConfig struct {
...
 	}
 
-	if c.WatermarkCritical == 0 {
-		c.WatermarkCritical = DefaultCriticalWatermark
+	if c.HighWatermark == 0 {
+		c.HighWatermark = DefaultHighWatermark
+	}
+
+	if c.LowWatermark > c.HighWatermark {
+		errors = append(errors, trace.BadParameter("low watermark (%v) cannot be higher than high watermark (%v)", c.LowWatermark, c.HighWatermark))
```
a-palchikov, line 71: This is going to be an issue unless you always set both values in a single go. Would it make sense to have defaults be relative to each other like default low watermark is always 10% smaller than the high watermark?

bernardjkim: https://github.com/gravitational/satellite/pull/224#discussion_r436213986
I think we wanted to be able to configure the two values separately.

Ok, so instead of returning an error if LowWatermark > HighWatermark, we'll just set LowWatermark = HighWatermark. This way LowWatermark <= HighWatermark.

monitoring/storage.go
```diff
@@ -50,9 +88,15 @@ type HighWatermarkCheckerData struct {
 	AvailableBytes uint64 `json:"available_bytes"`
 }
 
-// FailureMessage returns failure watermark check message
-func (d HighWatermarkCheckerData) FailureMessage() string {
-	return fmt.Sprintf("disk utilization on %s exceeds %v percent (%s is available out of %s), see https://gravitational.com/telekube/docs/cluster/#garbage-collection",
+// WarningMessage returns warning watermark check message
+func (d HighWatermarkCheckerData) WarningMessage() string {
+	return fmt.Sprintf("disk utilization on %s exceeds %v percent, cluster will degrade if usage exceeds %v percent (%s is available out of %s), see https://gravitational.com/gravity/docs/cluster/#garbage-collection",
```
r0mant, line 93: Nit: I'd move `(%s is available out of %s)` after the first part of the sentence, otherwise it reads a bit confusing:

```
disk utilization on %s exceeds %v percent (%s is available out of %s), cluster will degrade if usage exceeds %v percent, see https://gravitational.com/gravity/docs/cluster/#garbage-collection
```

Gopkg.toml
```diff
@@ -29,7 +29,7 @@ required = [
 
 [[constraint]]
   name = "github.com/gogo/protobuf"
-  version = "=v1.2.1"
+  version = "=v1.3.0"
```
r0mant, line 32: What's this dependency bump for? Just making sure we don't run into any incompatibility issues.

r0mant (approved):
Much cleaner now 👍

Gopkg.toml
bernardjkim: Just matching protobuf version in gravity https://github.com/gravitational/gravity/blob/95d4167587056e66cfecc55e391f4bff29d55481/Gopkg.toml#L74

a-palchikov (approved):

PR #226, opened 2020-06-08
Go modules

a-palchikov:
This PR migrates satellite to Go Modules.
Motivation is the bug in planet (which affects 7.1.x - i.e. master) which required careful plumbing around client-go's dependency on `golang.org/x/net` which if set incorrectly, resulted in http2 incompatibility down the road.
I had major difficulties with `mailgun/holser` - it would not work with modules due to a weird issue with `go-linq` module which I ended up abandoning and moved the `ttlmap` and related types into a [new](https://github.com/a-palchikov/ttlmap/) repository as I did not have commit permission for the original [repo](https://github.com/gravitational/ttlmap/).

Note also, that this PR is based off of profiling endpoint changes I've migrated from 5.5 branch which should be available as a separate PR.


go.mod
```diff
@@ -0,0 +1,55 @@
...
+	golang.org/x/sys v0.0.0-20200602225109-6fdc65e7d980
+	google.golang.org/grpc v1.23.0
+	gopkg.in/alecthomas/kingpin.v2 v2.2.6
+	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15
+	k8s.io/api v0.17.6
+	k8s.io/apimachinery v0.17.6
+	k8s.io/client-go v0.17.6
+)
+
+replace github.com/gravitational/ttlmap => github.com/a-palchikov/ttlmap v0.0.0-20200607134314-5bbd378b94a2
```
r0mant, line 55: Is this intended or a debug leftover?

a-palchikov: it is intended as temporary as long as I cannot commit to the original repository.

bernardjkim (approved):

go.mod
a-palchikov: figured it out.

mage.go
```diff
@@ -1,5 +1,3 @@
-// +build ignore
```
knisbet, line 1: I'm not sure why this was removed. My understanding, is that this sort of get's used as a convention, so that this file gets ignore my the build process and is only considered when running `go run mage.go` which will ignore the tag.

knisbet (approved):

mage.go
a-palchikov: I was just thinking about commenting on it - I ended up removing it because it was not letting me `go mod vendor` the mage [CLI](github.com/magefile/mage/mage) package dependency and failing at the test step with:
```
$ go run mage.go build:satellite
...
=====> Running GRPC Codegen inside docker...

mage.go:21:2: cannot find package "github.com/magefile/mage/mage" in any of:
	/go/src/github.com/gravitational/satellite/vendor/github.com/magefile/mage/mage (vendor tree)
	/usr/local/go/src/github.com/magefile/mage/mage (from $GOROOT)
	/go/src/github.com/magefile/mage/mage (from $GOPATH)
Error: running "docker run --volume=/home/deemok/go/src/github.com/gravitational/satellite:/go/src/github.com/gravitational/satellite:delegated satellite-grpc-buildbox:7.1.0-21-gb1ece2e9-dirty sh -c cd /go/src/github.com/gravitational/satellite/ && go run mage.go internal:grpcAgent internal:grpcDebug" failed with exit code 1
exit status 1
```
You can see the effect by undoing the change and running `go mod vendor` which then deletes the dependency:
```
modified:   mage.go
deleted:    vendor/github.com/magefile/mage/internal/run.go
deleted:    vendor/github.com/magefile/mage/mage/command_string.go
deleted:    vendor/github.com/magefile/mage/mage/magefile_tmpl.go
deleted:    vendor/github.com/magefile/mage/mage/main.go
deleted:    vendor/github.com/magefile/mage/mage/template.go
deleted:    vendor/github.com/magefile/mage/parse/parse.go
modified:   vendor/modules.txt
```
I'm not sure how to go about it yet as I cannot quite explain the behavior of `go mod` in this context.

a-palchikov: It looks like it only "handles" `ignore` tag in this way - if I use another one (i.e. `mage`), it does not happen - I'll change to use the new tag.

PR #229, opened 2020-06-11
[7.0.x] Update disk check

bernardjkim:
### Description
This PR updates the disk check to report two separate warning and critical probes at different thresholds.

### Linked tickets and PRs
Ports #224

### DIFF
--- a/monitoring/storage.go
+++ b/monitoring/storage.go
@@ -19,6 +19,8 @@ package monitoring
 import (
 	"fmt"
 
+	"github.com/gravitational/trace"
+
 	humanize "github.com/dustin/go-humanize"
 )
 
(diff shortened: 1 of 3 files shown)


monitoring/storage.go
```diff
@@ -50,17 +88,29 @@ type HighWatermarkCheckerData struct {
 	AvailableBytes uint64 `json:"available_bytes"`
 }
 
-// FailureMessage returns failure watermark check message
-func (d HighWatermarkCheckerData) FailureMessage() string {
-	return fmt.Sprintf("disk utilization on %s exceeds %v percent (%s is available out of %s), see https://gravitational.com/telekube/docs/cluster/#garbage-collection",
+// WarningMessage returns warning watermark check message
+func (d HighWatermarkCheckerData) WarningMessage() string {
+	return fmt.Sprintf("disk utilization on %s exceeds %v%% (%s is available out of %s), cluster will degrade if usage exceeds %v%%, see https://gravitational.com/gravity/docs/cluster/#garbage-collection",
```
knisbet, line 93: Hmm, I'm a bit on the fence on how much value it ads to include the output about what point the cluster will degrade at.

knisbet, line 93: Sort of in relation to above, it seems slightly inconsistent to report that we exceed a threshold of 80%, but then report usage in bytes and not %.

knisbet (approved):

monitoring/storage.go
bernardjkim: I think it's nice to be able to see how much disk is available in bytes, but yeah inconsistent with the percentage. We can include both, though the message is already quite long.

bernardjkim: Since this is just a user message we can leave it for now and change it later on if necessary. @a-palchikov suggested we include the point the cluster will degrade, so I think it adds some value to the user experience.

PR #233, opened 2020-06-15
[5.5.x] System pods checker

bernardjkim:
###  Description 
This PR adds a system pods checker. This checker verifies that all pods with the `gravitational.io/critical-pod` label is healthy.

### Linked tickets and PRs
* Ports https://github.com/gravitational/satellite/pull/186

### DIFF
--- a/monitoring/system_pods.go
+++ b/monitoring/system_pods.go
@@ -1,0 +1,249 @@
+/*
+Copyright 2020 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


monitoring/system_pods.go
```diff
@@ -0,0 +1,249 @@
...
+func verifyContainers(containerStatuses []corev1.ContainerStatus) error {
+	for _, status := range containerStatuses {
+		if status.State.Waiting != nil {
+			reason := status.State.Waiting.Reason
+			if reason == errImagePullBackOff || reason == errCrashLoopBackOff || reason == errImagePull {
+				return trace.BadParameter("%v waiting: %v", status.Name, status.State.Waiting.Reason)
+			}
+		}
+		if status.State.Terminated != nil {
+			reason := status.State.Terminated.Reason
```
a-palchikov, line 192: nit: I don't remember if it was discussed but what about exit code != 0 - has this been considered as a failing case as well?

a-palchikov (approved):

r0mant (approved):

monitoring/system_pods.go
bernardjkim: Hmm, checking if exit code != 0 probably works better here. Will update.

PR #241, opened 2020-06-29
Resolve dependency incompatibility

bernardjkim:
### Description
This modifies the serf and ttlmap dependencies to resolve incompatibility with teleport. 
* Serf v0.9.2 => v0.8.3
* ttlmap => ttlmap/v2

### Linked tickets and PRs
* Requires https://github.com/gravitational/ttlmap/pull/3


r0mant (approved):

agent/agent.go
```diff
@@ -219,15 +219,15 @@ func New(config *Config) (*agent, error) {
 	// Only initialize cluster timeline for master nodes.
 	var (
 		clusterTimeline history.Timeline
-		lastSeen        *ttlmap.TTLMap
+		lastSeen        *holster.TTLMap
```
a-palchikov, line 222: I'm not sure what triggered this change but I wasted a lot of time the last time I updated the dependencies for gravity (k8s v1.17.6) and this was a conscious change - getting rid of the holster (as we're only using the `TTLMap` and holster is somewhat bigger and has a problematic dependency on [go-linq](https://github.com/mailgun/holster/blob/v3.11.1/go.mod#L7)).
I ripped the TTLMap from holster into gravitational/ttlmap as part of [this](https://github.com/gravitational/ttlmap/pull/2) PR.

bernardjkim: @a-palchikov 
Teleport and Satellite use different versions of TTLMap (0.0.1 and master) so there is a conflict when trying to revendor satellite in gravity. I didn't want to change Teleport code so I just reverted back to holster in Satellite. If we'd like to remove holster dep, I'll just update satellite to use TTLMap v0.0.1 for now then.

PR #242, opened 2020-07-07
Edit time-drift check behavior

bernardjkim:
### Description
This PR slightly modifies the time drift check behavior. The time drift check can now report high time drift if there is an issue getting time drift from another node.

Also changed a few logs from warning to debug level because they were flooding the logs. 

### Purpose
While running through the test plan, I ran into the scenario where node-2 and node-3 were seeing high time drift and node-1's planet agent was stopped. node-2 and node-3 were unable to report high time-drift because the check was failing while trying trying to dial node-1's agent.

### DIFF
--- a/monitoring/timedrift.go
+++ b/monitoring/timedrift.go
@@ -136,7 +136,8 @@ func (c *timeDriftChecker) check(ctx context.Context, r health.Reporter) (err er
 	for _, node := range nodes {
 		drift, err := c.getTimeDrift(ctx, node)
 		if err != nil {
-			return trace.Wrap(err)
+			log.WithError(err).Debug("Failed to get time drift.")
+			continue
 		}
 		if isDriftHigh(drift) {
 			r.Add(c.failureProbe(node, drift))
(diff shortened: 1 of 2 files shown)


monitoring/timedrift.go
```diff
@@ -136,7 +136,8 @@ func (c *timeDriftChecker) check(ctx context.Context, r health.Reporter) (err er
 	for _, node := range nodes {
 		drift, err := c.getTimeDrift(ctx, node)
 		if err != nil {
-			return trace.Wrap(err)
+			log.WithError(err).Debug("Failed to get time drift.")
```
r0mant, line 139: Should this be logged at Error level then?

bernardjkim: I'd like to avoid filling the logs with this error as well. In the scenario mentioned above, this error would constantly be logged while node-1's agent is down.

I think in the cases when the checker would run into this error there should be other indication that there is an issue with the cluster. If a planet-agent is down, the node should show as offline, if tcp is dropped/rejected, the node should be offline as well.

r0mant: Okay, makes sense.

r0mant (approved):

a-palchikov (approved):

PR #244, opened 2020-07-09
Kernel checker

bernardjkim:
### Description
Add a kernel checker that checks if the currently installed kernel version is supported by Gravity.

### Linked tickets and PRs
* Updates https://github.com/gravitational/gravity/issues/1818
* Requires https://github.com/gravitational/satellite/pull/248


monitoring/boot_config_linux.go
```diff
@@ -110,31 +110,35 @@ func KernelVersionLessThan(version KernelVersion) KernelConstraintFunc {
 			(testVersion.Release == version.Release &&
```
a-palchikov, line 110:
```go
func KernelVersionLessThan(version KernelVersion) KernelConstraintFunc {
	return func(testVersion KernelVersion) bool {
		release := testVersion.Release <= version.Release
		major := testVersion.Major <= version.Major
		minor := testVersion.Minor <= version.Minor
		patch := testVersion.Patch < version.Patch
		return release && major && minor && patch
	}
}
```
otherwise `3.1.10-1125` < `3.0.10-1126`

monitoring/boot_config_linux.go
```diff
@@ -110,27 +110,35 @@ func KernelVersionLessThan(version KernelVersion) KernelConstraintFunc {
...
 	Release int
 	// Major specifies the major version component
 	Major int
 	// Minor specifies the minor version component
 	Minor int
+	// Patch specifies the patch or build number
+	Patch int
+}
+
+func (r *KernelVersion) String() string {
```
r0mant, line 140: Nit: Godoc.

monitoring/boot_config_linux.go
```diff
@@ -230,9 +238,10 @@ type kernelVersionReader func() (version string, err error)
 type bootConfigReader func(release string) (io.ReadCloser, error)
 
 func parseKernelVersion(input string) (*KernelVersion, error) {
```
r0mant, line 240: Could you add a comment on this function explaining what format of kernel version string it expects?

monitoring/boot_config_linux.go
```diff
@@ -253,10 +262,18 @@ func parseKernelVersion(input string) (*KernelVersion, error) {
 			"invalid kernel version minor: %v, expected a number",
 			parts[2])
 	}
+	patch, err := strconv.Atoi(parts[3])
+	if err != nil {
+		return nil, trace.BadParameter(
```
r0mant, line 267: I would add the entire input string (that's being parsed) to all these error messages.

monitoring/kernel.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (r *kernelChecker) check(_ context.Context, reporter health.Reporter) error {
+	release, err := realKernelVersionReader()
+	if err != nil {
+		return trace.Wrap(err, "failed to read kernel version")
+	}
+
+	kernelVersion, err := parseKernelVersion(release)
+	if err != nil {
+		reporter.Add(r.warningProbe(fmt.Sprintf("Failed to determine kernel version: %v.", release), err))
```
r0mant, line 72: Should this be a warning? I think we should only log this as an error and not create a probe.

monitoring/kernel.go
```diff
@@ -0,0 +1,104 @@
...
+	if !r.isSupportedVersion(*kernelVersion) {
+		reporter.Add(r.warningProbe(fmt.Sprintf("Installed linux kernel is not supported: %v.", release), err))
+	}
+	return nil
+}
+
+func (r *kernelChecker) successProbe() *pb.Probe {
+	return &pb.Probe{
+		Checker: r.Name(),
+		Detail:  "Installed linux kernel is supported.",
```
r0mant, line 85: Add kernel version to the message?

monitoring/kernel.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	kernelVersion, err := parseKernelVersion(release)
+	if err != nil {
+		reporter.Add(r.warningProbe(fmt.Sprintf("Failed to determine kernel version: %v.", release), err))
+		return trace.Wrap(err, "failed to determine kernel version")
+	}
+
+	if !r.isSupportedVersion(*kernelVersion) {
+		reporter.Add(r.warningProbe(fmt.Sprintf("Installed linux kernel is not supported: %v.", release), err))
```
r0mant, line 77: I'd make the message less scary. `Minimum recommended kernel version is %v (%v is installed).`

monitoring/kernel_test.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+		{
+			comment:  Commentf("Does not meet minor version requirement."),
+			expected: false,
+			version:  KernelVersion{Release: 3, Major: 10, Minor: 0},
+		},
+		{
+			comment:  Commentf("Does not meet patch number requirement."),
+			expected: false,
+			version:  KernelVersion{Release: 3, Major: 10, Minor: 0, Patch: 1126},
```
r0mant, line 58: Should probably be `Minor: 1` if you want to test patch component comparison.

monitoring/kernel_test.go
```diff
@@ -0,0 +1,68 @@
...
+		},
```
r0mant, line 59: Add a test case for 4.x.x kernel too?

monitoring/boot_config_linux.go
bernardjkim: Thanks for catching this. I've added a few extra test cases.

monitoring/kernel.go
```diff
@@ -0,0 +1,103 @@
+/*
```
a-palchikov, line 1: If we move kernel-specific functionality into a separate file and it's still relying on Linux-specific code (eg `realKernelVersionReader`) - the source file needs to be Linux-specific and suffixed with `_linux.go` (same for test file) - otherwise we're risking build failure during publish step on darwin.

r0mant (approved):

PR #254, opened 2020-08-12
[5.5.x] serf client

a-palchikov:
This PR attempts to rectify the issue we were seeing on clusters with longer runtime and possibly suspended (hibernated) nodes (VMs). I'll create an issue to describe the specific behavior and problem.
Here, instead of using a single client instance per lifetime of the agent, the client is only created when necessary so the connections to serf are short-lived. This will help with:
  * serf RPC implicitly client closing a connection upon encountering a certain condition which we attempted to address previously by having client check for underlying client validity and reconnecting if invalid
  * serf client failing with `i/o timeout` when the connection to serf has run into a state when the remote side does not respond anymore (although TCP keep-alive packets are still going).

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -586,7 +586,7 @@ func (r *AgentSuite) newAgent(config testAgentConfig, client *mockClusterMembers
 		localStatus:             config.localStatus,
 		lastSeen:                lastSeen,
 		statusQueryReplyTimeout: statusQueryReplyTimeout,
-		newSerfClient:           newClusterMembershipFrom(client),
+		newSerfClientFunc:       newClusterMembershipFrom(client),
 	}
 
 	if err := r.becomeActiveMember(client, agent); err != nil {
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -55,10 +55,8 @@ var _ = Suite(&AgentSuite{})
 
 func (r *AgentSuite) SetUpSuite(c *C) {
 	// Set logging level
-	if testing.Verbose() {
-		log.SetOutput(os.Stderr)
-		log.SetLevel(log.DebugLevel)
-	}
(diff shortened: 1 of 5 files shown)


agent/agent_test.go
```diff
@@ -723,18 +728,7 @@ func (r *AgentSuite) becomeActiveMember(membership *mockClusterMembership, agent
 	return r.becomeMember(membership, agent, MemberAlive)
 }
 
-// func (r *AgentSuite) becomeInactiveMember(membership *mockClusterMembership, agent *agent) error {
-// 	return r.becomeMember(membership, agent, MemberLeft)
-// }
-
 func (r *AgentSuite) becomeMember(membership *mockClusterMembership, agent *agent, status MemberStatus) error {
-	if _, ok := membership.members[agent.Name]; ok {
```
a-palchikov, line 731: Not sure I followed the logic here and why this check was in place? If the intention is to update the member's status (which I think was the primary goal), then it's irrelevant. Otherwise, I think it would be better to split creating an agent (`newAgent`) and adding a member (now `newAgent` creates an active member explicitly). @bernardjkim Let me know if I misunderstood this piece.

lib/membership/serf.go
```diff
@@ -30,35 +29,24 @@ import (
...
-		config: config,
 	}, nil
 }
 
 // Members lists members of the serf cluster.
-func (r *RetryingClient) Members() ([]ClusterMember, error) {
-	if err := r.reinit(); err != nil {
-		return nil, trace.Wrap(err)
-	}
-	r.RLock()
```
a-palchikov, line 60: No need to retry since connections are short-lived.

r0mant:
Going to close this one and open a different one instead so I can assign Dima as a reviewer since I've made some changes.

PR #255, opened 2020-08-19
(5.5) Update storage checker to support uid/gid checks

r0mant:
This pull request updates the existing "storage" checker to also support verifying expected path ownership (uid/gid). This will be used by planet/gravity to verify proper state directory ownership. Refs https://github.com/gravitational/gravity/issues/2022.

### DIFF
--- a/monitoring/storage.go
+++ b/monitoring/storage.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 Gravitational, Inc.
+Copyright 2018-2020 Gravitational, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -41,6 +41,10 @@ type StorageConfig struct {
 	// HighWatermark is the disk occupancy percentage that will trigger a critical probe.
 	// Disk usage check will be skipped if HighWatermark is unspecified or 0.
 	HighWatermark uint
+	// UID is the expected user owner of the path.
+	UID *uint32
+	// GID is the expected group owner of the path.
+	GID *uint32
 }
 
 // CheckAndSetDefaults validates this configuration object.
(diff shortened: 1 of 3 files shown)


bernardjkim (approved):

knisbet (approved):

monitoring/storage.go
```diff
@@ -101,3 +105,49 @@ func (d HighWatermarkCheckerData) SuccessMessage() string {
 
 // DiskSpaceCheckerID is the checker that checks disk space utilization
 const DiskSpaceCheckerID = "disk-space"
+
+// PathUIDCheckerData is attached to path UID check results.
+type PathUIDCheckerData struct {
```
a-palchikov, line 110: nit: for me, these two could be implemented as a single ID checker - would be simpler? They do not require implementation deviations beyond the semantics of the storage variable names (uid vs gid).

monitoring/storage_linux.go
```diff
@@ -147,6 +148,72 @@ func (c *storageChecker) checkFsType(ctx context.Context, reporter health.Report
...
+func (c *storageChecker) checkOwner(reporter health.Reporter) error {
+	if c.UID == nil && c.GID == nil {
+		return nil
+	}
+	uid, gid, err := c.osInterface.getUIDGID(c.Path)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if c.UID != nil {
+		data := PathUIDCheckerData{
```
a-palchikov, line 161: ... so these blocks could be reduced to something like:

```go
uid, gid, err := c.osInterface.getUIDGID(c.Path)
if err != nil {
	return trace.Wrap(err)
}
if c.UID != nil {
	c.verifyID(*c.UID, uid, reporter)
}
if c.GID != nil {
	c.verifyID(*c.GID, gid, reporter)
}
```

PR #262, opened 2020-08-28
(5.5) Serf client updates

r0mant:
This PR takes over the one started by @a-palchikov in https://github.com/gravitational/satellite/pull/254 with some additional changes which I'll describe below.

Refs https://github.com/gravitational/gravity/issues/2007.

## 1. Short-lived serf client

The main change that @a-palchikov implemented is refactoring the agent to establish the serf connection on-demand (on each health check cycle) rather than maintaining a long-lived connection. This part mostly remained as was implemented.

For testing of this one, since we don't know how it actually reproduces (even simulating timeouts with iptables doesn't seem to trigger the "freezes" we observed on the customer's clusters), all I did is stop serf on one of the nodes, make sure the agent fails to connect on each cycle and then start serf back and make sure the agent reconnects fine.

## 2. Crashed serf handling

This is something I've noticed while testing the above. If you stop serf on one node, the agent will happily return the last cached healthy local status to other clients so the node will remain "healthy" even though no checks are being run.

I'm not sure if that regressed at some point or has always been this way, but I've refactored the handling of failures during status loop to better handle such scenarios (e.g. failure to create serf client now that it's created dynamically) and now the cluster properly degrades:

```
dod-cluster-node-3:/$ planet status --local
{
   "name": "10_128_0_34.test",
   "member_status": {
      "name": "10_128_0_34.test"
   }
}[ERROR]: status degraded
```
```
dod-cluster-node-3:/$ planet status
{
   "status": "degraded",
   "timestamp": "2020-08-28T00:10:06.494254171Z",
   "summary": "failed to create serf client: failed to connect to serf agent: client.Config{Addr:\"127.0.0.1:7373\", AuthKey:\"\", Timeout:0}"
}[ERROR]: status degraded
```

## 3. Unnecessary serf tags updates

When creating a serf client, we would also immediately issue a request to update serf member tags which I found out is actually redundant since we already set these tags on the command-line (by passing `--tag` flag to serf agent) so that call was really not doing anything useful and only generating extra requests. This would be esp. unnecessary since the client is re-created every cycle now.

So I've removed that call from `newSerfClient()` method and made sure all serf members still have proper tags:

```
$ serf members
10_128_0_33.test  10.128.0.33:7496  alive   role=master,publicip=10.128.0.33
10_128_0_34.test  10.128.0.34:7496  failed  publicip=10.128.0.34,role=master
10_128_0_32.test  10.128.0.32:7496  alive   role=master,publicip=10.128.0.32
```

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -551,12 +551,8 @@ func (r *agent) defaultUnknownStatus() *pb.NodeStatus {
 	return &pb.NodeStatus{
 		Name: r.Name,
 		MemberStatus: &pb.MemberStatus{
-			Name:   r.Name,
-			Addr:   r.RPCAddrs[0],
-			Status: pb.MemberStatus_None,
-			Tags:   r.Tags,
+			Name: r.Name,
 		},
-		Status: pb.NodeStatus_Unknown,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


knisbet (approved):

agent/agent.go
```diff
@@ -553,34 +540,54 @@ func (r *agent) statusUpdateLoop(ctx context.Context) error {
...
+		MemberStatus: &pb.MemberStatus{
+			Name: r.Name,
+		},
+	}
+}
+
 // collectStatus obtains the cluster status by querying statuses of
 // known cluster members.
-func (r *agent) collectStatus(ctx context.Context) (systemStatus *pb.SystemStatus, err error) {
+func (r *agent) collectStatus(ctx context.Context) pb.SystemStatus {
```
a-palchikov, line 561: nit: I'd return the pointer since this is how the return is expected.

a-palchikov (approved):

bernardjkim (approved):

PR #265, opened 2020-09-11
scaling: timedrift false positives at scale

knisbet:
This PR implements a couple of changes to the timedrift check to support cluster scaling:
- Parallel RPC calls to peers
- Discard of results from slow peers that may produce inaccurate results
- Delta against the time the response was received, not when it was sent (RPC calls may cold start, need to TCP connect, etc).

Updates https://github.com/gravitational/gravity/issues/2121

### DIFF
--- a/monitoring/timedrift.go
+++ b/monitoring/timedrift.go
@@ -40,6 +40,14 @@ const (
 	// timeDriftThreshold sets the default threshold of the acceptable time
 	// difference between nodes.
 	timeDriftThreshold = 300 * time.Millisecond
+
+	// timeDriftCheckTimeout drops time checks where the RPC call to the remote server take too long to respond.
+	// If the client or server is busy and the request takes time to be processed, this will cause an innaccurate
+	// comparison of the current time.
+	timeDriftCheckTimeout = 100 * time.Millisecond
+
+	// parallelRoutines indicates how many parallel queries we should run to peer nodes
+	parallelRoutines = 20
 )
 
 // timeDriftChecker is a checker that verifies that the time difference between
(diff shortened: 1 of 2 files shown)


monitoring/timedrift.go
```diff
@@ -40,6 +40,14 @@ const (
 	// timeDriftThreshold sets the default threshold of the acceptable time
 	// difference between nodes.
 	timeDriftThreshold = 300 * time.Millisecond
+
+	// timeDriftCheckTimeout drops time checks where the RPC call to the remote server take too long to respond.
+	// If the client or server is busy and the request takes time to be processed, this will cause an innaccurate
```
r0mant, line 45: innaccurate -> inaccurate

monitoring/timedrift.go
```diff
@@ -182,6 +217,11 @@ func (c *timeDriftChecker) getTimeDrift(ctx context.Context, node serf.Member) (
 	// Obtain this node's local timestamp.
 	t1Start := c.Clock.Now().UTC()
 
+	// if the RPC call takes a long duration it will result in an innaccurate comparison. Timeout the RPC
```
r0mant, line 220: innaccurate -> inaccurate

monitoring/timedrift.go
```diff
@@ -203,7 +243,7 @@ func (c *timeDriftChecker) getTimeDrift(ctx context.Context, node serf.Member) (
 	// Finally calculate the time drift between this and the specified node
 	// using formula: T2 - T1Start - Latency.
```
r0mant, line 244: The comment needs updating too then.

r0mant (approved):

monitoring/timedrift.go
knisbet: It's actually a good catch, I made a small semantic difference in the formula. I've now switched it to `now - t2 - latency` such that now - t2 should be a positive duration, and the -latency should adjust back to estimate time due to latency.

monitoring/timedrift.go
```diff
@@ -133,16 +141,43 @@ func (c *timeDriftChecker) check(ctx context.Context, r health.Reporter) (err er
...
+
+	var wg sync.WaitGroup
+
+	wg.Add(parallelRoutines)
+
+	for i := 0; i < parallelRoutines; i++ {
+		go func() {
+			for {
+				select {
+				case node := <-nodesC:
```
a-palchikov, line 160: I think it can be simpler:
```go
nodesC := make(chan serf.Member, len(nodes))
for _, node := range nodes {
	nodesC <- node
}
close(nodesC)

var wg sync.WaitGroup
var mutex sync.Mutex
wg.Add(parallelRoutines)
for i := 0; i < parallelRoutines; i++ {
	go func() {
		for node := range nodesC {
			drift, err := c.getTimeDrift(ctx, node)
			if err != nil {
				log.WithError(err).Debug("Failed to get time drift.")
				continue
			}
			if isDriftHigh(drift) {
				mutex.Lock()
				r.Add(c.failureProbe(node, drift))
				mutex.Unlock()
			}
		}
		wg.Done()
	}()
}
```

monitoring/timedrift_test.go
```diff
@@ -126,7 +140,7 @@ func (s *TimeDriftSuite) TestTimeDriftChecker(c *check.C) {
 				Clock:      s.clock,
 			},
 			FieldLogger: logrus.WithField(trace.Component, "test"),
-			clients:     newClientsCache(test.times),
+			clients:     newClientsCache(test.times, test.slow),
```
a-palchikov, line 143: Why not just use an expired context? You can create one in the SetUpSuite or as a package global:
```go
s.ctx, cancel = context.WithCancel(context.Background())
cancel()
```
and then just use it in the test instead of `slow` - this will require minimum changes.

a-palchikov (approved):

PR #267, opened 2020-09-22
[master] scale tuning

knisbet:
1. Instead of locating the nethealth pod by querying the kubernetes cluster, nethealth will instead listen to a unix domain socket that will be mapped to planet. Satellite queries the metrics via unix domain socket.
    Note: I left the prometheus endpoint, just incase there is a use case still for querying over the network
2. Do a bit of cleanup on the timedrift calculations. Makes it a bit easier for me atleast to understand the algorithm, of which I think I reversed the +/- a few times on the adjustment factor for the query latency.
3. Have system pod checkers run only on master nodes (this is changed within planet). This way, 800 workers aren't all trying to query kubernetes to check just themselves.

### DIFF
--- a/lib/nethealth/nethealth.go
+++ b/lib/nethealth/nethealth.go
@@ -81,6 +81,9 @@ const (
 )
 
 type Config struct {
+	// PrometheusSocket is the path to a unix socket that can be used to retrieve the prometheus metrics
+	PrometheusSocket string
+
 	// PrometheusPort is the port to bind to for serving prometheus metrics
 	PrometheusPort uint32
 
@@ -244,20 +247,24 @@ func (s *Server) Start() error {
 			s.Fatalf("ListenAndServe(): %s", err)
 		}
 	}()
-	go func() {
-		_ = os.Remove(DefaultNethealthSocket)
 
-		unixListener, err := net.Listen("unix", DefaultNethealthSocket)
+	if s.config.PrometheusSocket != "" {
+		_ = os.Remove(s.config.PrometheusSocket)
+
+		unixListener, err := net.Listen("unix", s.config.PrometheusSocket)
 		if err != nil {
-			panic(err)
+			return trace.Wrap(err)
 		}
 
-		if err := s.httpServer.Serve(unixListener); err != http.ErrServerClosed {
-			s.Fatalf("Unix Listen(): %s", err)
-		}
-	}()
+		go func() {
+			if err := s.httpServer.Serve(unixListener); err != http.ErrServerClosed {
+				s.Fatalf("Unix Listen(): %s", err)
+			}
+		}()
+	}
 
 	s.Info("Started nethealth with config:")
+	s.Info("  PrometheusSocker: ", s.config.PrometheusPort)
 	s.Info("  PrometheusPort: ", s.config.PrometheusPort)
 	s.Info("  Namespace: ", s.config.Namespace)
 	s.Info("  NodeName: ", s.config.NodeName)
--- a/lib/nethealth/nethealth.go
+++ b/lib/nethealth/nethealth.go
@@ -240,6 +240,16 @@ func (s *Server) Start() error {
 			s.Fatalf("ListenAndServe(): %s", err)
 		}
 	}()
+	go func() {
+		unixListener, err := net.Listen("unix", "/run/nethealth/nethealth.sock")
+		if err != nil {
+			panic(err)
+		}
+
+		if err := s.httpServer.Serve(unixListener); err != http.ErrServerClosed {
+			s.Fatalf("Unix Listen(): %s", err)
+		}
+	}()
 
 	s.Info("Started nethealth with config:")
 	s.Info("  PrometheusPort: ", s.config.PrometheusPort)
--- a/lib/nethealth/nethealth.go
+++ b/lib/nethealth/nethealth.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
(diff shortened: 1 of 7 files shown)


lib/nethealth/nethealth.go
```diff
@@ -240,6 +244,18 @@ func (s *Server) Start() error {
 			s.Fatalf("ListenAndServe(): %s", err)
 		}
 	}()
+	go func() {
+		_ = os.Remove(DefaultNethealthSocket)
```
a-palchikov, line 248: L248-L255 (everything up to httpServer.Serve should be executed outside of the goroutine as it does not make sense to continue upon these failures. This will eliminate the need to panic

monitoring/nethealth.go
```diff
@@ -331,12 +296,14 @@ func nethealthFailureProbe(name, peer string, packetLoss float64) *pb.Probe {
 
 // fetchNethealthMetrics collects the network metrics from the nethealth pod
 // specified by addr. Returns the resp as an array of bytes.
-func fetchNethealthMetrics(ctx context.Context, addr string) ([]byte, error) {
-	client, err := roundtrip.NewClient(addr, "")
-	if err != nil {
-		return nil, trace.Wrap(err, "failed to connect to nethealth service at %s.", addr)
+func (c *nethealthChecker) fetchNethealthMetrics(ctx context.Context) (res []byte, err error) {
```
a-palchikov, line 299: nit: update the comment to remove the mention of removed `addr`

monitoring/system_pods.go
```diff
@@ -29,13 +29,10 @@ import (
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 )
 
 // SystemPodsConfig specifies configuration for a system pods checker.
 type SystemPodsConfig struct {
-	// NodeName specifies the kubernetes name of this node.
-	NodeName string
```
bernardjkim, line 38: Hmm, we're removing the `nodeName` field selector so that master nodes will also check the status of pods running on non-master nodes? That makes sense.
One concern is we will end up with duplicate probes on all master nodes with this change.

monitoring/nethealth.go
```diff
@@ -331,12 +296,14 @@ func nethealthFailureProbe(name, peer string, packetLoss float64) *pb.Probe {
...
+	client := http.Client{
+		Transport: &http.Transport{
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", nethealth.DefaultNethealthSocket)
```
a-palchikov, line 303: instead of using the package global, I'd have nethealthChecker has the socket path as configuration and set it to `DefaultNethealthSocket` if unspecified.

monitoring/nethealth.go
```diff
@@ -349,11 +316,27 @@ func fetchNethealthMetrics(ctx context.Context, addr string) ([]byte, error) {
 	//      # TYPE nethealth_echo_timeout_total counter
 	//      nethealth_echo_timeout_total{node_name="10.128.0.96",peer_name="10.128.0.70"} 37
 	//      nethealth_echo_timeout_total{node_name="10.128.0.96",peer_name="10.128.0.97"} 0
-	resp, err := client.Get(ctx, client.Endpoint("metrics"), url.Values{})
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "http://unix/metrics", nil)
```
a-palchikov, line 319: nit: do we have all clients using go 1.13 where it was introduced?

monitoring/system_pods.go
knisbet: Correct, we don't currently have a way to deduplicate checks. Also, it's sort of the case for other checks like nethealth or time sync. All masters will report that. 

Yes it should be better, but it's within the realm of the current design.

monitoring/nethealth.go
knisbet: Depends if someone decides to backport. AFAIK 7.0 on gravity/planet are built with go 1.13, which are the only intended backport targets at present.

monitoring/system_pods.go
knisbet: Ideally we'd also be able to have a master node return a result for a different node, so the status would ultimately be displayed under say a worker node. I might add that later, but for this change I need to work with what we have.

monitoring/nethealth.go
```diff
@@ -329,14 +294,16 @@ func nethealthFailureProbe(name, peer string, packetLoss float64) *pb.Probe {
...
 }
 
-// fetchNethealthMetrics collects the network metrics from the nethealth pod
-// specified by addr. Returns the resp as an array of bytes.
-func fetchNethealthMetrics(ctx context.Context, addr string) ([]byte, error) {
-	client, err := roundtrip.NewClient(addr, "")
-	if err != nil {
-		return nil, trace.Wrap(err, "failed to connect to nethealth service at %s.", addr)
+// fetchNethealthMetrics collects the network metrics from the nethealth pod.
+// Returns the resp as an array of bytes.
```
a-palchikov, line 298: nit: resp -> response

a-palchikov:
Looks like I failed to submit the couple of comments previously although I was sort of certain I did :-(

knisbet:
> Looks like I failed to submit the couple of comments previously although I was sort of certain I did :-(

I just threw it into a new PR and merged since I already merged this one: https://github.com/gravitational/satellite/pull/268

PR #271, opened 2020-09-28
[7.0.x] fix http context

knisbet:
- This fixes an issue on backport, where we're using a version of golang that doesn't have http.NewRequestWithContext
- Also fixes an issue I noticed in the tests, where the tests panic if `testing.Verbose` is called within `init()`. I just dropped the logging config within the test suite for now, there might be a better way to address this long term.

### DIFF
--- a/monitoring/nethealth.go
+++ b/monitoring/nethealth.go
@@ -303,7 +303,6 @@ func (c *nethealthChecker) fetchNethealthMetrics(ctx context.Context) (res []byt
 				return net.Dial("unix", c.NethealthSocketPath)
 			},
 		},
-		Timeout: time.Second,
 	}
 	// The two relevant metrics exposed by nethealth are 'nethealth_echo_request_total' and
 	// 'nethealth_echo_timeout_total'. We expect a pair of request/timeout metrics per peer.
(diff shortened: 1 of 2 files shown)


monitoring/nethealth.go
```diff
@@ -303,7 +303,6 @@ func (c *nethealthChecker) fetchNethealthMetrics(ctx context.Context) (res []byt
 				return net.Dial("unix", c.NethealthSocketPath)
```
a-palchikov: I think this is inferior to using the net.Dialer directly (and thus making good use of the context):
```go
DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
	var d net.Dialer
	return d.DialContext(ctx, "unix", c.NethealthSocketPath)
}
```

monitoring/nethealth.go
```diff
@@ -303,7 +303,6 @@ func (c *nethealthChecker) fetchNethealthMetrics(ctx context.Context) (res []byt
...
 			},
 		},
-		Timeout: time.Second,
```
a-palchikov: Not sure why this is removed - usually, it makes sense to limit the http client to avoid unrestricted connect attempts. So using either timeouts or context would be preferred.

a-palchikov (approved):

PR #278, opened 2020-10-16
Update system pods check

bernardjkim:
## Description
This PR modifies the system pods check to query only the kube-system and monitoring namespaces. This will help  reduce the load on etcd if there are a large number of pods in user namespaces.

### DIFF
--- a/lib/kubernetes/constants.go
+++ b/lib/kubernetes/constants.go
@@ -19,10 +19,4 @@ package kubernetes
 const (
 	// AllNamespaces can be used to query pods in all namespaces.
 	AllNamespaces = ""
-
-	// NamespaceKubeSystem specifies kube-system namespace.
-	NamespaceKubeSystem = "kube-system"
-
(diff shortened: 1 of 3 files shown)


lib/kubernetes/constants.go
```diff
@@ -16,5 +16,13 @@ limitations under the License.
...
-const AllNamespaces = ""
+const (
+	// AllNamespaces can be used to query pods in all namespaces.
+	AllNamespaces = ""
+
+	// NamespaceKubeSystem specifies kube-system namespace.
+	NamespaceKubeSystem = "kube-system"
+
+	// NamespaceMonitoring specifies monitoring namespace.
+	NamespaceMonitoring = "monitoring"
```
r0mant, line 27: I don't think we should hardcode namespaces for satellite to check. Let planet-agent pass them to satellite via checker configuration.

bernardjkim: Updated in 7049874

r0mant (approved):

PR #279, opened 2020-10-16
[5.5.x] Update system pods check

bernardjkim:
## Description
This PR modifies the system pods check to query only select namespaces for critical system pods. This will help reduce the load on etcd if there are a large number of pods in user namespaces.

## Linked tickets and PRs
* Ports https://github.com/gravitational/satellite/pull/278

### DIFF
--- a/monitoring/system_pods.go
+++ b/monitoring/system_pods.go
@@ -22,7 +22,6 @@ import (
 
 	"github.com/gravitational/satellite/agent/health"
 	pb "github.com/gravitational/satellite/agent/proto/agentpb"
-	"github.com/gravitational/satellite/lib/kubernetes"
 	"github.com/gravitational/satellite/utils"
 
 	"github.com/gravitational/trace"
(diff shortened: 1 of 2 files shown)


r0mant (approved):

monitoring/system_pods.go
```diff
@@ -104,17 +105,21 @@ func (r *systemPodsChecker) check(ctx context.Context, reporter health.Reporter)
...
+func (r *systemPodsChecker) getPods() (pods []corev1.Pod, err error) {
 	opts := metav1.ListOptions{
 		LabelSelector: systemPodsSelector.String(),
 		FieldSelector: fields.OneTermEqualSelector("spec.nodeName", r.NodeName).String(),
 	}
-	pods, err := r.Client.CoreV1().Pods(kubernetes.AllNamespaces).List(opts)
-	if err != nil {
-		return nil, utils.ConvertError(err)
+
+	for _, namespace := range r.Namespaces {
```
a-palchikov, line 114: Is AllNamespaces still supposed to be used as a fallback for empty configuration (`len(r.Namespaces) == 0`)?

bernardjkim: I think it would be better to just skip the check if an empty configuration is provided. If we default to AllNamespaces we may risk users experiencing the issue that this PR was trying to solve.

PR #282, opened 2020-10-20
Update cluster membership interface

bernardjkim:
## Description
This PR updates the cluster membership interface. Cluster interface is no longer intended to manage the cluster membership, its only purpose is to query membership status. `IsMember` and `Join` methods have been removed from the agent interface. Management of cluster membership should be handled outside of satellite and in planet or gravity. 

Updates the time drift checker and latency checker to now interface with the cluster membership interface.

This is an intermediate PR for removing Serf dependency from satellite. Serf cluster implementation will be replaced with an implementation using Kubernetes api.

### DIFF
--- a/agent/proto/agentpb/status.go
+++ b/agent/proto/agentpb/status.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2016 Gravitational, Inc.
+Copyright 2016-2020 Gravitational, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -22,6 +22,16 @@ import "github.com/gravitational/trace"
 // EmptyStatus returns an empty system status
 func EmptyStatus() *SystemStatus {
 	return &SystemStatus{Status: SystemStatus_Unknown}
+}
+
+// NewMemberStatus constructs a new MemberStatus.
+func NewMemberStatus(name, addr string, tags map[string]string) *MemberStatus {
+	return &MemberStatus{
+		Name:   name,
+		Addr:   addr,
+		Status: MemberStatus_Alive,
+		Tags:   tags,
+	}
 }
 
 // encoding.TextMarshaler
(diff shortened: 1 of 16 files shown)


a-palchikov (approved):

agent/proto/agentpb/status.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2016 Gravitational, Inc.
+Copyright 2016-2020 Gravitational, Inc.
```
lenko-d, line 2: I think that the date should show a single date(2020) instead of a range. Same for the other places.

lenko-d (approved):

agent/proto/agentpb/status.go
bernardjkim: I have been updating the copyright notice dates this way up til now because I thought this was the correct way. But after reading up a bit on copyright notices, it seems that copyright notices don't necessarily need to be included in the header in the first place. There doesn't seem to be a right or wrong way to date the notices, so I guess we just need consistency.

@r0mant @knisbet not that important of an issue, but do we have a preferred way of dating the copyright notices?

knisbet: We use a date range the same way as was done in the PR, it's fine. We're not very consistent about setting it, would be nice to script it at some point but not high on the list of priorities.

A couple examples:
- https://github.com/gravitational/gravity/blob/master/tool/gravity/cli/register.go#L2
- https://github.com/gravitational/gravity/blob/master/lib/ops/ops.go#L2
-

knisbet (approved):

monitoring/latency/latency_test.go
```diff
@@ -85,33 +84,26 @@ func (r *LatencySuite) TestLatency(c *C) {
...
-				Spec: v1.PodSpec{
-					NodeName: node2,
-				},
-			},
-		},
-	},
-)
+// mockCluster implements a mock cluster to be used for testing.
+type mockCluster struct {
+	membership.Cluster
```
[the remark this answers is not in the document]
bernardjkim: Yup, not sure if this follows best practices, but the latency checker currently never makes a call to `Member()` so the Cluster is just embedded to satisfy the interface.

PR #283, opened 2020-10-27
OpenEBS enhancements

lenko-d:
This PR provides the following OpenEBS related enhancements:

* A new checker, that checks that the iscsid is not running on the host when
 OpenEBS is enabled in the Gravity's app deployment manifest. This is needed because if iscsid is running on the host it
 makes the iscsid in Gravity's Planet to fail. 

Also this helps to avoid possible issues with version mismatch between iscsid on the host and what is expected by OpenEBS in the deployment. If the versions don't match the app will not work when deployed on a host with incompatible iscsid. Installing iscsid in Gravity's Planet avoids the version mismatch.

### DIFF
--- a/monitoring/defaults_linux.go
+++ b/monitoring/defaults_linux.go
@@ -18,6 +18,24 @@ package monitoring
 
 import "github.com/gravitational/satellite/agent/health"
 
+var (
+	// DefaultProcessesToCheck is the default list of processs to verify that are not running on the host.
+	// This list should be passed to the DefaultProcessChecker function.
+	DefaultProcessesToCheck = []string{
+		"dockerd",
+		"docker-current", // Docker daemon name when installed from RHEL repos.
+		"lxd",
+		"coredns",
+		"kube-apiserver",
+		"kube-scheduler",
+		"kube-controller-manager",
+		"kube-proxy",
+		"kubelet",
+		"planet",
+		"teleport",
+	}
+)
+
 // DefaultPortChecker returns a port range checker with a default set of port ranges
 func DefaultPortChecker() health.Checker {
 	return NewPortChecker(
@@ -51,31 +69,13 @@ func PreInstallPortChecker() health.Checker {
 }
 
 // DefaultProcessChecker returns checker which will ensure no conflicting program is running
-func DefaultProcessChecker(openEBSEnabled bool) health.Checker {
-	processes := []string{
-		"dockerd",
-		"docker-current", // Docker daemon name when installed from RHEL repos.
-		"lxd",
-		"coredns",
-		"kube-apiserver",
-		"kube-scheduler",
-		"kube-controller-manager",
-		"kube-proxy",
-		"kubelet",
-		"planet",
-		"teleport",
-	}
-
-	if openEBSEnabled {
-		processes = append(processes, "iscsid")
-	}
-
-	return &ProcessChecker{processes}
+func DefaultProcessChecker(processesToCheck []string) health.Checker {
+	return &ProcessChecker{processesToCheck}
 }
 
 // BasicCheckers detects common problems preventing k8s cluster from
 // functioning properly
-func BasicCheckers(openEBSEnabled bool, checkers ...health.Checker) health.Checker {
+func BasicCheckers(processesToCheck []string, checkers ...health.Checker) health.Checker {
 	c := &compositeChecker{
 		name: "local",
 		checkers: []health.Checker{
@@ -86,7 +86,7 @@ func BasicCheckers(openEBSEnabled bool, checkers ...health.Checker) health.Check
 			NewWormholeWgForwardingChecker(),
 			NewBridgeNetfilterChecker(),
 			NewMayDetachMountsChecker(),
-			DefaultProcessChecker(openEBSEnabled),
+			DefaultProcessChecker(processesToCheck),
 			DefaultPortChecker(),
 			DefaultBootConfigParams(),
 			NewInotifyChecker(),
@@ -97,8 +97,8 @@ func BasicCheckers(openEBSEnabled bool, checkers ...health.Checker) health.Check
 }
 
 // PreInstallCheckers are designed to run on a node before installing telekube
-func PreInstallCheckers(openEBSEnabled bool) health.Checker {
-	return BasicCheckers(openEBSEnabled, PreInstallPortChecker())
+func PreInstallCheckers(processesToCheck []string) health.Checker {
+	return BasicCheckers(processesToCheck, PreInstallPortChecker())
 }
 
 // DefaultBootConfigParams returns standard kernel configs required for running kubernetes
--- a/monitoring/defaults_linux.go
+++ b/monitoring/defaults_linux.go
@@ -31,8 +31,6 @@ var (
 		"kube-controller-manager",
 		"kube-proxy",
 		"kubelet",
-		"planet",
-		"teleport",
 	}
 )
 
--- a/monitoring/defaults_linux.go
+++ b/monitoring/defaults_linux.go
@@ -51,8 +51,8 @@ func PreInstallPortChecker() health.Checker {
 }
 
 // DefaultProcessChecker returns checker which will ensure no conflicting program is running
-func DefaultProcessChecker() health.Checker {
-	return &ProcessChecker{[]string{
+func DefaultProcessChecker(openEBSEnabled bool) health.Checker {
+	processes := []string{
 		"dockerd",
 		"docker-current", // Docker daemon name when installed from RHEL repos.
 		"lxd",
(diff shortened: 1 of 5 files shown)


monitoring/defaults_linux.go
```diff
@@ -51,8 +51,8 @@ func PreInstallPortChecker() health.Checker {
 }
 
 // DefaultProcessChecker returns checker which will ensure no conflicting program is running
-func DefaultProcessChecker() health.Checker {
-	return &ProcessChecker{[]string{
+func DefaultProcessChecker(openEBSEnabled bool) health.Checker {
```
r0mant, line 54: I don't think satellite should be aware of OpenEBS specifically as it's intended to be a general purpose monitoring tool. What I would probably do, is extend it with a list of additional processes to check and pass `iscsid` from Gravity i.e. something like:

```go
func DefaultProcessChecker(additionalProcesses ...string) health.Checker { ... }
```

monitoring/defaults_linux.go
```diff
@@ -64,12 +64,18 @@ func DefaultProcessChecker() health.Checker {
...
+		processes = append(processes, "iscsid")
+	}
+
+	return &ProcessChecker{processes}
 }
 
 // BasicCheckers detects common problems preventing k8s cluster from
 // functioning properly
-func BasicCheckers(checkers ...health.Checker) health.Checker {
+func BasicCheckers(openEBSEnabled bool, checkers ...health.Checker) health.Checker {
```
r0mant, line 78: Same comment as above, make it agnostic to OpenEBS, just make the public API more flexible.

monitoring/iscsi.go
```diff
@@ -0,0 +1,83 @@
...
+	return iscsiCheckerID
+}
+
+// Check will check the systemd unit data coming from dbus to verify that
+// there are no iscsid services present on the host.
+func (c iscsiChecker) Check(ctx context.Context, reporter health.Reporter) {
+	conn, err := dbus.New()
+	if err != nil {
+		reason := "failed to connect to dbus"
+		reporter.Add(NewProbeFromErr(c.Name(), reason, trace.Wrap(err)))
```
r0mant, line 55: Is this missing a `return`?

monitoring/iscsi.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	defer conn.Close()
+
+	units, err := conn.ListUnits()
+	if err != nil {
+		reason := "failed to query systemd units"
+		reporter.Add(NewProbeFromErr(c.Name(), reason, trace.Wrap(err)))
```
r0mant, line 62: `return`?

monitoring/iscsi.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	probeFailed := false
+	for _, unit := range units {
+		if unit.Name == "iscsid.service" || unit.Name == "iscsid.socket" {
```
r0mant, line 67: You could do 
```go
switch unit.Name {
case "iscisid.service", "iscsid.socket":
  ...
}
```

which is a bit more idiomatic for Go.

monitoring/iscsi.go
```diff
@@ -0,0 +1,83 @@
...
+			if unit.ActiveState == "active" || unit.LoadState != loadStateMasked {
+				reporter.Add(&pb.Probe{
+					Checker: iscsiCheckerID,
+					Detail: fmt.Sprintf("Found conflicting systemd service: %v. "+
```
r0mant, line 71: This message will be confusing to a user, i.e. I think it should explain why the service is conflicting.

monitoring/iscsi.go
```diff
@@ -0,0 +1,83 @@
...
+	units, err := conn.ListUnits()
+	if err != nil {
+		reason := "failed to query systemd units"
+		reporter.Add(NewProbeFromErr(c.Name(), reason, trace.Wrap(err)))
+	}
+
+	probeFailed := false
+	for _, unit := range units {
+		if unit.Name == "iscsid.service" || unit.Name == "iscsid.socket" {
+			if unit.ActiveState == "active" || unit.LoadState != loadStateMasked {
```
r0mant, line 68: Nit: Let's be consistent and use either constants or raw strings (`"active"` vs `loadStateMasked`).

monitoring/iscsi.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Name returns the name of this checker
+// Implements health.Checker
+func (c iscsiChecker) Name() string {
+	return iscsiCheckerID
+}
+
+// Check will check the systemd unit data coming from dbus to verify that
+// there are no iscsid services present on the host.
+func (c iscsiChecker) Check(ctx context.Context, reporter health.Reporter) {
```
r0mant, line 51: Please write a unit test for this checker.

monitoring/defaults_linux.go
a-palchikov: Strongly agree - default presets do not work most of the time. I'm not sure how many of these are use in gravity, but I would definitely make the defaults exported (e.g. the list of processes) and add the ability to set the configuration externally so it would be possible to configure tests without touching the satellite checker code.

monitoring/iscsi.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, unit := range units {
+		if unit.Name == "iscsid.service" || unit.Name == "iscsid.socket" {
+			if unit.ActiveState == "active" || unit.LoadState != loadStateMasked {
+				reporter.Add(&pb.Probe{
+					Checker: iscsiCheckerID,
+					Detail: fmt.Sprintf("Found conflicting systemd service: %v. "+
+						"Please stop, disable and mask this service and try again.", unit.Name),
+					Status: pb.Probe_Failed,
+				})
+				probeFailed = true
```
a-palchikov, line 75: You don't need this boolean - just return early and always add a successful probe after making sure every test passed (i.e. at the very end of Check).

lenko-d: > You don't need this boolean - just return early and always add a successful probe after making sure every test passed (i.e. at the very end of Check).

If I don't have a boolean and return early after the first unit has failed then I will not be showing all unit failures to the user. There can be 2 unit checks failing at the same time and I want to show them all at once.

a-palchikov: You could query `reporter.NumProbes()` and use it for the same purpose.

monitoring/iscsi_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/gravitational/satellite/agent/health"
+	pb "github.com/gravitational/satellite/agent/proto/agentpb"
+	"github.com/gravitational/satellite/lib/test"
+
+	"github.com/coreos/go-systemd/v22/dbus"
+	. "gopkg.in/check.v1"
+)
+
+const (
+	ActiveAndMasked   = "ActiveState is inactive and LoadState is masked."
```
a-palchikov, line 31: nit: these need not be exported.
Also, `ActiveAndMasked` is actually `InactiveAndMasked`

a-palchikov (approved):

monitoring/iscsi.go
```diff
@@ -0,0 +1,95 @@
...
+const (
+	iscsiCheckerID = "iscsi"
+
+	// ISCSIDService is the name of the iscsid systemd service unit.
+	ISCSIDService = "iscsid.service"
+	// ISCSIDSocket is the name of the iscsid systemd socket unit.
+	ISCSIDSocket = "iscsid.socket"
+)
+
+type FormatISCSIError func(unitName string) string
```
bernardjkim, line 38: nit: Add godocs for exported type.

monitoring/iscsi.go
```diff
@@ -0,0 +1,95 @@
...
+// Name returns the name of this checker
+// Implements health.Checker
+func (c iscsiChecker) Name() string {
+	return iscsiCheckerID
+}
+
+// Check will check the systemd unit data coming from dbus to verify that
+// there are no iscsid services present on the host.
+func (c iscsiChecker) Check(ctx context.Context, reporter health.Reporter) {
+	conn, err := dbus.New()
```
bernardjkim, line 59: Recommend just logging dbus connection error as it does not necessarily verify iscsid service status. Same below with `conn.ListUnits()` error.

I think a false positive is better than a false negative here since this checker reports a failed probe and will degrade the cluster.

bernardjkim (approved):

monitoring/iscsi_test.go
```diff
@@ -0,0 +1,119 @@
...
+			comment:    Commentf(inactiveAndMasked),
+			unitStatus: []dbus.UnitStatus{{Name: ISCSIDService, ActiveState: activeStateInactive, LoadState: loadStateMasked}},
+			probe: &pb.Probe{
+				Checker: iscsiCheckerID,
+				Detail:  "",
+				Status:  pb.Probe_Running,
+			},
+		},
+		{
+			comment:    Commentf(inactiveAndMasked),
```
knisbet, line 103: nit: I think it's better to use unique comments per table test, so it's alot clearer if one of these starts failing exactly which test case to look at.

r0mant (approved):

monitoring/iscsi.go
lenko-d: I think that this is correct behavior. If we cannot query the status of the systemd units via dbus then this means that something major is not ok with the system and the cluster deserves to be in a degraded state. 

We cannot say that the cluster is ok if we cannot query the state of the systemd units via dbus.  Dbus and systemd are very important components and if we cannot query them then we should definitely declare the cluster as degraded.

wadells:
We're unlikely to merge this PR as the primary author has moved on, and the motivation is not as pressing with gravity's deprecation.  If we revisit the ideas herein, we'll either reopen or move the patch stack to a new PR.

PR #284, opened 2020-11-16
Replace serfmembership with k8smembership

bernardjkim:
## Description
This PR replaces the cluster membership implementation with one that relies directly on Kubernetes. Satellite no longer relies on a Serf cluster to query cluster membership, instead all Nodes in the Kubernetes cluster will be considered a member of the cluster.

Will work on completely removing Serf packages and references in a later PR.

## Linked tickets and PRs
* Requires https://github.com/gravitational/satellite/pull/282


lib/membership/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,137 @@
...
+		c.Assert(cache.WaitForCacheSync(stop, informer.HasSynced), Equals, true, comment)
+
+		cluster, err := NewCluster(&Config{
+			Informer: informer,
+			Stop:     stop,
+		})
+		c.Assert(err, IsNil, comment)
+
+		members, err := cluster.Members()
+		sort.Sort(pb.ByName(members))
```
a-palchikov, line 79: Any reason why you sort the result before checking the error?

lib/membership/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for _, tc := range tests {
+		comment := Commentf(tc.comment)
+
+		factory := informers.NewSharedInformerFactory(fake.NewSimpleClientset(&v1.NodeList{Items: tc.nodes}), 0)
+		informer := factory.Core().V1().Nodes().Informer()
+
+		stop := make(chan struct{})
+		defer close(stop)
+		go informer.Run(stop)
```
a-palchikov, line 68: nit: this is going to get called only upon exit from the test method - have you considered confining the actual test in a separate function. In case more test cases are added, this might have a side-effect.
On the other hand, it looks like there's just a single test-case - why use table-driven testing prematurely?

lib/membership/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// Config contains information needed to identify other members deployed in
+// the Kubernetes cluster.
+type Config struct {
+	// Informer specifies a Node Infomer.
+	Informer cache.SharedIndexInformer
+	// Stop specifies a stop channel that indicates that the informer has been
+	// stopped when it is closed.
+	Stop <-chan struct{}
```
a-palchikov, line 46: Is this used for anything?

a-palchikov (approved):

lenko-d (approved):

lib/membership/kubernetes/kubernetes_test.go
bernardjkim: No reason. Will change to call sort after checking the error.

lib/membership/kubernetes/kubernetes_test.go
bernardjkim: Good point. I guess there is no need to use table-driven testing here.

PR #290, opened 2020-12-28
[BUGFIX] [WIP] fix Latency Check behavior

eldios:
As mentioned in:
  https://github.com/gravitational/gravity/issues/2067

Fixing the LatencyCheck behavior by lowering percentile and introducing minimum threshold of datapoints collected.


wadells (approved):
Min data points LGTM. In gravitational/gravity#2067 Roman talked about 20 being low.  What is your logic behind 10?

PR #299, opened 2021-03-11
Upgrade k8s client lib to 1.19.8

bernardjkim:
## Description
This PR upgrades the k8s client lib to 1.19.8. Includes necessary changes to modified API. These changes mainly include providing a context and metav1 options to each call.


monitoring/interpod.go
```diff
@@ -229,28 +229,50 @@ func waitTimeoutForPodRunningInNamespace(client *kube.Clientset, podName string,
...
+		"pod":       podName,
+		"namespace": ns,
+	})
+
+	log.Infof("waiting up to %v for pod status to be %s", timeout, desc)
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	ticker := time.NewTicker(pollInterval)
```
bernardjkim, line 242: Just realizing now, that after initialization of a ticker, nothing is received on the channel until after the specified time duration. 

In this case it is probably not a big deal since `pollInterval` is only 2 seconds, so the function will not wait too long before initial execution of the embedded code block.

In places where a longer interval is used, the embedded code block might end up waiting quite a bit before initial execution.

I think we use this pattern quite often when implementing retry or a continuous loop. Might be areas for optimization.

monitoring/interpod.go
```diff
@@ -229,28 +229,50 @@ func waitTimeoutForPodRunningInNamespace(client *kube.Clientset, podName string,
...
+			}
+
+			done, err := condition(pod)
+			if !done {
+				log.Infof("waiting for pod status to be '%s' (found phase: %q, readiness: %t) (%v elapsed)",
+					desc, pod.Status.Phase, podReady(pod), time.Since(start))
+				continue
+			}
+
 			if err != nil {
```
r0mant, line 264: Wouldn't the code will basically never reach inside this "err != nil" block because presumably if err != nil than "done" will be false and thus it will always do "continue" above? Maybe you want to check "err != nil" first and then "!done"?

r0mant (approved):

monitoring/interpod.go
bernardjkim: I just kept the same code paths as before and didn't bother to look at the implementation of `condition` 😅 , but it looks like the only time `err != nil` is when `done == true`. On line 218:
```golang
        return waitForPodCondition(client, namespace, podName, "running", timeout, func(pod *v1.Pod) (bool, error) {
		if pod.Status.Phase == v1.PodRunning {
			log.Infof("found pod '%s' on node '%s'", podName, pod.Spec.NodeName)
			return true, nil
		}
		if pod.Status.Phase == v1.PodFailed {
			return true, trace.Errorf("pod in failed status: %s", fmt.Sprintf("%#v", pod))
		}
		return false, nil
	})
```

monitoring/interpod.go
knisbet: Yea, most areas this pattern is used does overlook the first run. In most cases it doesn't make a huge difference since it doesn't matter too much if things take a few extra seconds, but is good to keep in mind.

knisbet (approved):

wadells (approved):

PR #300, opened 2021-03-12
Refactor gRPC codegen

bernardjkim:
## Description
Slight refactor of gRPC codegen. PR also bumps `protobuf` version which will be required when upgrading to `client-go` v1.19.


build.go
```diff
@@ -44,9 +44,8 @@ var (
 	// buildVersion allows override of the version string from env variable
 	buildVersion = env("BUILD_VERSION", "")
 
-	grpcProtocVersion = env("GRPC_PROTOC_VER", "3.7.1")      // version of protoc
-	grpcGogoProtoTag  = env("GRPC_GOGO_PROTO_TAG", "v1.2.1") // gogo version tag
-	grpcGatewayTag    = env("GRPC_GATEWAY_TAG", "v1.8.5")    // grpc gateway version tag
+	grpcProtocVersion = env("GRPC_PROTOC_VER", "3.14.0")     // version of protoc
+	grpcGogoProtoTag  = env("GRPC_GOGO_PROTO_TAG", "v1.3.1") // gogo version tag
```
a-palchikov, line 48: I believe latest version of gogo is `v1.3.2` - why not use it?

build.assets/grpc/Dockerfile
```diff
@@ -12,26 +12,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.10-stretch
+# Documentation on getting started with protobuf:
+# https://github.com/gogo/protobuf#getting-started
+
+# Requires 1.9+
+FROM golang:1.16-buster
```
a-palchikov, line 19: Does this include Go 1.16.2 security patch https://groups.google.com/g/golang-announce/c/ZWvSr9XM4wM/m/t8wy_q5_AgAJ?

build.assets/grpc/Dockerfile
```diff
@@ -12,26 +12,32 @@
...
-ENV PROTO_INCLUDE "/usr/local/include":"${GOPATH}/src":"${GOPATH}/src/github.com/gogo/protobuf/protobuf":"${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis":"${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis":"${GOGOPROTO_ROOT}:${GOGOPROTO_ROOT}/protobuf":"${GOGOPROTO_ROOT}/gogoproto"
+# Install and extract standard protocol buffer implementation from
+# https://github.com/protocolbuffers/protobuf
+# Standard .proto files will be extracted under ${PROTOC_PATH}/include.
+RUN curl -L -o /tmp/${TARBALL} https://github.com/protocolbuffers/protobuf/releases/download/v${GRPC_PROTOC_VER}/${TARBALL}
+RUN cd /tmp && unzip /tmp/${TARBALL} -d ${PROTOC_PATH} && rm /tmp/${TARBALL}
 
+# These binaries enable use of gogoprotobuf extensions
+# https://github.com/gogo/protobuf/blob/master/extensions.md
+RUN go get -u github.com/gogo/protobuf/proto@${GRPC_GOGO_PROTO_TAG}
```
a-palchikov, line 42: See [release notes](https://golang.org/doc/go1.16#go-command) Go 1.16 regarding changes in behavior of `go get` - `go install` with `@latest` would be more appropriate.
How does latest version however is handled since `gogo` is specified with a tag? Does the build script handle the use of the correct `gogo` version?

build.go
bernardjkim: client-go `v1.19` is using gogo `v1.3.1`. I figured it'd be safer to just keep the versions the same. https://github.com/kubernetes/client-go/blob/e9b0f31f935be48b6589f6026a2cb1891d92fe12/go.mod#L13

build.assets/grpc/Dockerfile
bernardjkim: Looks like this tag is kept up to date with the latest minor release. https://hub.docker.com/_/golang

build.assets/grpc/Dockerfile
bernardjkim: Thanks for pointing this out. I didn't realize changes were made to go get/install. I've made changes in 2b2b84d. However, it looks like we do not install `../proto`. The src files just need to be included when using `protoc`.

Build script handles use of correct `gogo` version.

wadells (approved):
LGTM!

knisbet (approved):

PR #305, opened 2022-05-27
[7.0.x] Run etcd-healthz checks in parallel

bernardjkim:
This PR modifies the `etcd-healthz` checks to run concurrently. This is an attempt to address https://github.com/gravitational/gravity/issues/2740.  I'm guessing for larger clusters, the `etcd-healthz` checks may be timing out because they are running sequentially. This change will allow the checks to run 10 at a time.

### DIFF
--- a/monitoring/composite.go
+++ b/monitoring/composite.go
@@ -18,8 +18,10 @@ package monitoring
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


wadells (approved):
LGTM.

PR #306, opened 2022-12-01
 Create initial codeql.yml configuration

jentfoo:
An initial configuration of CodeQL for `go`.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,40 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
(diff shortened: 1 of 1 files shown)


adaadb6 (approved):

bpdohall:
Do we want to add scanning to this repo? 

It is part of [gravity](https://github.com/gravitational/gravity) which is subject to deprecation (with distribution discontinuing in 6 months). 

Any findings from CodeQL will not get resolved.

bpdohall (approved):

jentfoo:
@bpdohall Thank you for that context.  I was trying to be thorough and did not realize this was being deprecated.  I agree this is not needed given the deprecation timeline, thank you!

